PR #1, opened 2025-05-30
Modify derivation to support wormhole

vrom911:
Additionally, I moved the `get_test_vectors` into a separate module as it was slowing down my editor (too big) and fmt didn't work well on that


hdwallet/src/lib.rs
```diff
@@ -79,8 +82,27 @@ impl HDLattice {
...
             return Ok(self.master_key);
         }
-        let entries = path.split('/');
+
+        // Check if this is a wormhole path
+        let (is_wormhole, remaining_path) = if path.starts_with("w/") {
+            (true, &path[2..])
+        } else {
+            (false, path)
+        };
```
n13, line 91: Feels like we should either chop off the first character for both cases or for none?

illuzen: let's do that in a separate pr tho

vrom911:
> We have some code to do it in chain, maybe we should pull that in here?

I think this would make sense to have it all in one place here

illuzen:
LGTM let's bump the version and merge

PR #2, opened 2025-06-12
Libp2p compatibility

illuzen:
These changes are to make our library compatible with libp2p


dilithium/src/dilithium5.rs
```diff
@@ -1,3 +1,7 @@
+use crate::errors::KeyParsingError;
```
n13, line 1: Should we not delete all of these except 87?

n13 (approved):
Good changes, but we should have a cleanup PR removing all the dilithum versions we don't use... I think all except ml_dsa_87

PR #3, opened 2025-06-12
chore: Monorepo layout

czareko:


n13:
So just renamed the directories?

PR #6, opened 2025-07-03
expose inner hash on wormhole pair

illuzen:
this will be used to make miners use wormholes

also this simplifies the verify api


n13 (approved):
Cool

PR #8, opened 2025-07-18
Switched out bip39 provider to be bip44 compliant

illuzen:
- We were parsing paths ourselves and in the process deviated from the bip44 standard (intentionally). But to facilitate integration with existing wallets like Ledger, we want to support bip44 paths. We use the nam-tiny-hderive library, which is very small and easy to audit. 
- We explicitly error when non-hardened paths are attempted
- We had to regenerate test vectors here, which means old seed phrases will now generate different keys. Obviously this would be a big issue if we were on mainnet, which is why we are doing it now.
- We make a public constant for quantus chain id, but we don't enforce it in the dilithium case
- We do wormhole addresses thru a different chain id, instead of w prefix, as that is not bip44 compliant


hdwallet/src/lib.rs
```diff
@@ -67,102 +81,54 @@ impl HDLattice {
         })
     }
 
-    pub fn master_key_from_seed(seed: &[u8; 64]) -> Result<[u8; 64], HDLatticeError> {
-        let mut hasher = hmac::Hmac::<Sha512>::new_from_slice(SALT).map_err(|_| {
-            HDLatticeError::KeyDerivationFailed("Failed to create HMAC".to_string())
-        })?;
-        hasher.update(seed);
-        Ok(hasher.finalize_fixed().into())
+    pub fn master_key_from_seed(seed: &[u8; 64]) -> Result<[u8; 32], HDLatticeError> {
```
n13, line 84: Is this a standard? To just use "m"?

So our initial master key is then also different - it's the same algo, but they're using a different salt. 

This means I can't use this to find our old keys - which is fine, I can just add this as a compatibility layer in the wallet.

n13 (approved):
LGTM

One comment - would be amazing to have some externally generated key as test vector to prove we're compatible with ... stuff. 

I checked the code they use "Bitcoin seed" as salt - I wonder if that's a standard? 

Also side-note - any clients, like the wallet, that want to support our old keys, will just have to re-implement the original master hash method. It's OK though, this is cleaner. And all wallets going forward will use it.

PR #13, opened 2025-08-28
Use stable toolchain and update plonky2 version

dastansam:


illuzen (changes requested):
let's use main branch on github dependencies. i know we are migrating to crates.io soon...

PR #50, opened 2026-01-09
Threshold signatures

illuzen:
Threshold ML-DSA-87 Signature Scheme

This PR adds a new `threshold` crate implementing threshold ML-DSA-87 (Dilithium) signatures for the NEAR MPC network.

### Overview

In a (t, n) threshold scheme:
- **n** total parties hold key shares
- Any **t** or more parties can cooperate to produce a valid signature  
- Fewer than **t** parties cannot sign or learn the secret key

Signatures produced are fully compatible with standard ML-DSA-87 verification.

### Features

- **ML-DSA-87**: NIST Level 5 post-quantum security (~256-bit)
- **Flexible Thresholds**: Supports (t, n) configurations where 2 ≤ t ≤ n ≤ 7
- **4-Round Signing Protocol**: Commitment, reveal, response, and leader decision phases
- **Distributed Key Generation (DKG)**: 4-round protocol to generate keys without a trusted dealer
- **Key Resharing**: Transfer keys to a new committee while preserving the public key
- **Key Derivation**: HD-wallet style derived keys via DKG contribution derivation
- **Leader-Based Retry**: Automatic retry coordination on rejection sampling failures
- **Message Buffering**: Handles out-of-order network messages gracefully
- **NEAR MPC Compatible**: Follows the poke/message `Protocol` trait pattern

### Technical Highlights

- **Replicated Secret Sharing (RSS)**: Each party holds shares for multiple subsets, enabling flexible threshold signing
- **Hyperball Sampling**: Ensures combined responses stay within rejection bounds with high probability
- **K Parallel Iterations**: Runs multiple signing attempts simultaneously to reduce retry rates
- **Serialization**: Uses serde + bincode for protocol message serialization (enabled by default)

### Supported Configurations

| Parties (n) | Thresholds (t) |
|-------------|----------------|
| 2 | 2 |
| 3 | 2, 3 |
| 4 | 2, 3, 4 |
| 5 | 2, 3, 4, 5 |
| 6 | 2, 3, 4, 5, 6 |
| 7 | 2, 3, 4, 5, 6, 7 |

### Testing

- 146 unit tests
- 19 integration tests covering all configurations
- 26 resharing tests
- 9 derivation tests
- Benchmarks for keygen, DKG, signing, and verification

### Other Changes

- Updated license to GPL-3.0
- Minor dependency updates in hdwallet and tests crates


n13:
This should probably become a new PR , since this one is quite messy - can be from same branch

it needs to be updated to current master status though... 

AI review: 

# PR #50 Review — Threshold Signatures

I checked out the branch (`illuzen/threshold`), built the workspace (`cargo check --workspace`), ran clippy (4 warnings), and ran the test suite (147 lib tests + 19 integration tests, all green). Then I read through the bulk of the threshold crate and skimmed the non-threshold changes. This is a serious, well-organised piece of work; the feedback below is in rough priority order.

## Blockers

### 1. Branch is badly out of date vs `master` — non-threshold changes are effectively a revert

The diff vs `master` includes (unintentional?) regressions of recently merged work:

- `hdwallet/src/hderive.rs`: `"Dilithium seed"` → `"Bitcoin seed"`, `SensitiveBytes32` → `k256::SecretKey` + `Protected`, reintroduces `k256`, `bs58`, and a `quantus_v1` feature with non-hardened paths. See diff in `hderive.rs` around lines ≈2205–2440 of the full diff.
- `hdwallet/src/wormhole.rs`: reverts to the old `hash_bytes/hash_twice/bytes_to_digest/string_to_felts` API and re-adds `Result<WormholePair, WormholeError>` to `generate_new`.
- `hdwallet/src/test_vectors.rs`: entirely regenerated (~10k lines).
- `dilithium/src/ml_dsa_87.rs`, `sign.rs`, `stack_usage_demo.rs`: `rand::rng()` → `rand::thread_rng()`.
- `Cargo.toml` workspace deps: `qp-poseidon-core` `=1.4.0` → `=1.0.7`, `qp-rusty-crystals-dilithium` `2.4.0` → `2.1.0`, `qp-rusty-crystals-hdwallet` `2.3.1` → `1.3.0`. `dilithium/Cargo.toml` and `hdwallet/Cargo.toml` follow the downgrade.

Master has commits like `new wormhole api`, `bump qp-poseidon to 1.4`, `bump dilithium to 2.4.0`, `bump hdwallet to 2.3.1` that this branch merged past (commits `9da6dbb`, `5ec263a`, `4277d5d`) but didn’t absorb into its modifications. The workspace only compiles because the downgraded versions happen to resolve; once merged this will silently revert all of those PRs.

**This needs a real rebase/merge against `master` before it can land.** The bip44 test vectors, wormhole API, k256-based derivation, and version numbers all need to be reconciled.

### 2. Real bug: `n = 7` signing uses wrong hyperball radius

`ThresholdConfig::new` accepts `(t, 7)` combinations and has K values for all of them:

```79:85:threshold/src/config.rs
// n = 7 (EXPERIMENTAL - not from reference implementation)
(2, 7) => 7,
(3, 7) => 55,
(4, 7) => 160,
(5, 7) => 320,
(6, 7) => 270,
(7, 7) => 650,
```

But `get_threshold_params` in the signing protocol only matches up to `(6, 6)` and falls through to the 2-of-2 default:

```355:372:threshold/src/protocol/signing.rs
match (config.threshold(), config.total_parties()) {
    (2, 2) => (503119.0, 503192.0, 7.0),
    // ...
    (6, 6) => (424124.0, 424197.0, 7.0),
    _ => (503119.0, 503192.0, 7.0), // Default to 2-of-2 values
}
```

So any n=7 signing session (and any unknown `(t,n)` that slipped through validation) silently uses 2-of-2 `r`/`r'` radii. That will either produce unsigned/rejected attempts forever or, worse, produce signatures whose rejection-sampling bound doesn’t match the claimed security level. Either:
- reject `n=7` at `ThresholdConfig::new` until tuned radii exist, or
- add the correct `(r, r', nu)` tuples and remove the catch-all default so unknown configs are a hard error (matches the user rule “Fail early — never have a silent failure”).

Ideally `k_iterations`, `(r, r', nu)`, and the sharing-pattern maximum are all co-located on `ThresholdConfig` so they can never drift apart.

## High-severity issues

### 3. Cross-platform non-determinism from `f64::ln / sin / cos`

`HyperballSampleVector::sample_hyperball` uses `f1.ln()`, `(2π·f2).cos()`, and `(2π·f2).sin()`:

```199:214:threshold/src/protocol/primitives.rs
let z1 = (-2.0 * f1.ln()).sqrt() * (2.0 * PI * f2).cos();
let z2 = (-2.0 * f1.ln()).sqrt() * (2.0 * PI * f2).sin();
```

These are not guaranteed to be bit-identical across platforms/libm versions (x86 vs ARM, glibc vs musl vs macOS). The comment “matching Golang exactly” is about the uniform conversion but the transcendentals are platform libm. If two NEAR MPC nodes diverge in the last ULP of `ln`, their `round()` of `y_k` can flip, `w_k` will differ, and every downstream commitment hash will mismatch (HQ2 check fails). In a distributed protocol this is a liveness time-bomb.

Options: switch to a deterministic integer sampler (e.g. CBD / sampled-from-shake integers), pin a software libm (`libm` crate), or explicitly test cross-platform with CI on at least linux x86_64, linux aarch64, and macOS.

### 4. Silent error swallowing in protocol paths

Several places violate the “fail early / no silent failures” rule, which for a threshold protocol also has security implications (an adversary sending garbage can get itself quietly excluded from aggregation rather than hard-failing):

- `signer.rs` drops parties whose Round 3 unpack fails:

```481:486:threshold/src/signer.rs
for r3 in all_round3 {
    if r3.party_id != self.private_key.party_id() {
        if let Ok(responses) = unpack_responses(&r3.response, &self.config) {
            all_responses.push(responses);
        }
    }
}
```

Same pattern for `combine_with_message` at line ≈533, and `aggregate_commitments_dilithium`/`unpack_commitment_dilithium` inside `round3_respond` at `signer.rs:410-419`. A malformed broadcast from party B results in party B being silently omitted from aggregation; at best this just fails signing, at worst it hides an attack. Return `ThresholdError::InvalidSignatureShare { party_id, .. }`/`InvalidCommitment` instead.

- DKG `eprintln!` + discarded errors:

```241-243, 253-255, 286-288, 298-300, 351-352, 366-367, 389-390, 399-400, 414-416, 474-476, 512-514:threshold/src/keygen/dkg/protocol.rs
eprintln!("Error processing buffered message: {}", e);
...
eprintln!("Failed to deserialize message from {}: {}", from, e);
...
eprintln!("Failed to process message from {}: {}", from, e);
```

For a library `eprintln!` is the wrong sink (consumers can’t control it) and the error itself is discarded — `message()` returns `()` and `poke()` continues. Use `log`/`tracing` and/or return an error so the caller can react. Otherwise a peer sending malformed bincode can DoS the round silently.

- `let _ = self.state_data.transition_to(...)` (lines 209, 216, 447) discards transition errors, while the same function’s result is iterated in sibling branches. Either both paths should be treated or the function’s signature should change.

### 5. Non-deterministic HashMap iteration drives network sends

In DKG the P2P round-4 messages are built and indexed by iterating `HashMap<u16, SubsetContribution>`:

```747-775, 800-831:threshold/src/keygen/dkg/protocol.rs
for (subset_mask, contribution) in &secret_contributions.subset_contributions { ... }
...
for (subset_mask, _) in &secret_contributions.subset_contributions { ... } // matched by msg_index
```

This relies on the fact that HashMap iteration for a single instance within one process is stable. It is — but (a) it’s fragile under refactors, (b) clippy already flags the second form, and (c) if the two loops ever diverge (e.g. a rehash insertion between them) recipients and contents de-sync silently. Use `BTreeMap<u16, _>` for anything that drives protocol ordering, or precompute `Vec<(ParticipantId, DkgRound4Private)>` once and index it.

### 6. `HyperballSampleVector` witness data is not zeroized

`Round1Data::zeroize()` only `.clear()`s `w_commitments` and `hyperball_samples`:

```44-52:threshold/src/protocol/signing.rs
self.commitment_hash.zeroize();
self.rho_prime.zeroize();
self.w_commitments.clear();
self.hyperball_samples.clear();
```

`Vec::clear` drops the elements, but `HyperballSampleVector::data: Box<[f64]>` is not zeroized on drop. Those floats are the secret witness `y` (plus `cs2 + y`) — compromising them reveals `s1`. Implement `Zeroize`/`ZeroizeOnDrop` on `HyperballSampleVector` and do the same for the `Vec<polyvec::Polyvecl>` witnesses in `SignerState::AfterRound3`. The manual `for i in 0..7 { resp.vec[i].coeffs.fill(0); }` in `signer::reset` also hard-codes L=7 — use `L` from constants.

## Medium-severity issues

### 7. DRY: duplicated modular reduction in `dealer.rs`

`threshold/src/keygen/dealer.rs` lines 363-381 re-define `reduce_le2q`, `le2q_mod_q`, `mod_q` already present in `threshold/src/protocol/primitives.rs`. Import them — otherwise a future fix/optimization in one copy won’t propagate. (Also directly flagged by the user’s DRY rule.)

### 8. Leader-based Round 4 — trust model needs calling out

`DilithiumSignProtocol` has a single `leader_id` that unilaterally decides `Round4Complete(sig)` vs `Round4Retry`. Followers accept the signature without independently running `combine_signature` (verification is outside the protocol). For NEAR MPC with a trusted committee this is fine, but:
- The README / module doc should state the trust assumption explicitly (“leader must not lie; otherwise the worst case is denial-of-signature / extra retries”).
- Consider having followers independently verify the bytes in `Round4Complete` via `verify_signature(pk, msg, ctx, sig)` before returning `Action::Return(sig)`. Cheap, and defends against a byzantine leader quietly burning retry rounds.

### 9. `ThresholdSigner::new` has correct subset-signing checks, but missing `MAX_PARTIES` check

`generate_with_dealer` does not explicitly enforce `config.total_parties() <= MAX_PARTIES`; it only relies on `ThresholdConfig::new` having been called first. `ThresholdConfig` is `Copy + Clone` and gets passed around freely, so the invariant is fine today, but add a `debug_assert!` in `generate_with_dealer`/`dealer.rs` to be defensive against future refactors.

### 10. `ThresholdConfig::deserialize` silently re-derives `k_iterations`

```131-144:threshold/src/config.rs
let data = ConfigData::deserialize(deserializer)?;
ThresholdConfig::new(data.threshold, data.total_parties).map_err(serde::de::Error::custom)
```

Good — but the serialized field is called `ConfigData` with only 2 fields, while `Serialize` writes a struct with 2 fields named `threshold`/`total_parties`. If anyone ever adds a field to the struct, the round-trip breaks silently. A `#[derive(Serialize, Deserialize)]` on a wire-only type with explicit reconstruction via `new()` is fine; just worth a comment.

## Low-severity / nits

- `lib.rs:94` + signer docs call it a “3-round” protocol in one place and a “4-round” (including leader decision) elsewhere. Pick one phrasing in `README.md` / `lib.rs` / `signing_protocol.rs`.
- `keygen/dkg/mod.rs` top comment says “5-round DKG” but the `DkgState` enum and `Action` flow actually walk 5 generating phases + waiting phases; accurate but inconsistent with the “4-round DKG” claim in the PR body.
- `HyperballSampleVector` has a custom `pub fn clone(&self) -> Self` instead of `#[derive(Clone)]`. Derive it so it satisfies `Clone` trait bounds.
- `SignerState::default()` impl is written before `SignerState` is declared (`signer.rs:96-121`). Works because Rust item order is free, but reads oddly.
- `RESHARING_PLAN.md` is a planning doc — should it live in-tree or move to an issue/wiki after landing?
- `threshold/papers/ThresholdMLDSA.pdf` (~750 KB) is committed. Consider referencing an arxiv/IACR URL instead of shipping the PDF inside the crate.
- Dozens of `if let Ok(...)` / `if end <= data.len()` silent-skip patterns in `protocol/signing.rs::unpack_responses` and `signer.rs`. Collectively same category as #4.
- Clippy still reports 4 easy fixes (`unnecessary_lazy_evaluations` in `signer.rs:387`, `for_kv_map` in `dkg/protocol.rs:811`, `unwrap_or_default` in `dkg/state.rs:371`, `needless_range_loop` in `dkg/mod.rs:475`, `doc_lazy_continuation` in `protocol/signing.rs:351`). Worth clearing before merge.

## Nice things

- Security posture in the threshold crate is genuinely solid: `#![deny(unsafe_code)]`, `ZeroizeOnDrop` on `PrivateKeyShare`/`ThresholdSigner`, `Debug` redacts key bytes (`keys.rs:227-237`), the HQ1–HQ7 fixes have dedicated tests (`test_hq1_fix_threshold_property_preserved` in particular is a nice regression test).
- Lib test coverage (147) and integration tests (19) covering every valid `(t, n)` from `(2,2)` up to `(5,7)` including DKG + subset-signing paths.
- `participants::ParticipantList` with id→index mapping is a clean abstraction for NEAR’s large `u32` ids, mirroring threshold-signatures lib.
- Protocol state machines (`SignerState`, `DkgState`, `SignProtocolState`) are explicit enums with clear transitions; message buffering for out-of-order delivery is thoughtful.

## Recommendation

Don’t merge as-is. The two blockers (#1 outdated master, #2 silent n=7 default) are mandatory fixes. Items #3 (float determinism) and #4/5 (silent error swallowing / HashMap ordering) are the items I’d want a maintainer response on before signing off — they affect correctness in the adversarial/distributed setting this crate exists to handle. The rest is polish.

Happy to dig into any of these in more depth — e.g. I can prototype a fix for #2 or a `Zeroize` impl for `HyperballSampleVector` (#6), or propose a `BTreeMap` refactor for #5 — just say the word.

PR #62, opened 2026-03-07
Tests from NIST testing tool

hdvanegasm:
This commit contains tests from the NIST testing tool. I took only the "free" test which covers more than 300+ test cases.

This is just an informal testing and should not be considered a complete testing using the full NIST tool.


illuzen:
thanks, merged into https://github.com/Quantus-Network/qp-rusty-crystals/pull/78

PR #74, opened 2026-05-02
Threshold Signatures - 2 

illuzen:
## Summary

Adds the `qp-rusty-crystals-threshold` crate implementing threshold ML-DSA-87 (Dilithium) signatures for the NEAR MPC network, along with security improvements to the dilithium crate.

## Threshold ML-DSA-87

This crate enables multiple parties to collectively sign messages without any single party having access to the complete signing key. In a (t, n) threshold scheme:
- **n** total parties hold key shares
- Any **t** or more parties can cooperate to produce a valid signature  
- Fewer than **t** parties cannot sign or learn the secret key

Signatures are fully compatible with standard ML-DSA-87 verification.

### Features

- **ML-DSA-87**: NIST Level 5 post-quantum security (~256-bit)
- **Flexible Thresholds**: Supports (t, n) configurations where 2 ≤ t ≤ n ≤ 7
- **4-Round Signing Protocol**: Commitment, reveal, response, and leader decision phases
- **Leader-Based Retry**: Automatic retry on rejection sampling failures
- **Distributed Key Generation (DKG)**: 4-round Mithril protocol to generate keys without a trusted dealer
- **Key Resharing**: Transfer keys to a new committee while preserving the public key
- **Key Derivation**: HD-wallet style derived keys for NEAR MPC integration
- **Message Buffering**: Handles out-of-order network messages
- **`no_std` Compatible**: Core functionality works without standard library

### Key Concepts

- **Replicated Secret Sharing (RSS)**: Each party holds shares for multiple subsets of signers
- **Hyperball Sampling**: Ensures combined randomness stays within rejection bounds
- **K Iterations**: Runs K parallel signing attempts to minimize retry probability

### Supported Configurations

| Parties (n) | Thresholds (t) |
|-------------|----------------|
| 2 | 2 |
| 3 | 2, 3 |
| 4 | 2, 3, 4 |
| 5 | 2, 3, 4, 5 |
| 6 | 2, 3, 4, 5, 6 |
| 7 | 2, 3, 4, 5, 6, 7 |

### Dilithium Crate
- **KeccakState**: Added `Zeroize` and `ZeroizeOnDrop` traits, removed `Copy` derive to prevent accidental copies of sensitive cryptographic state
- **SecretKey**: Made internal `bytes` field private to prevent accidental exposure of secret key material
- **SHAKE256 Fix**: Fixed bug in `keccak_squeeze` where multi-block squeezes (>136 bytes) would overwrite data

## Testing

All tests passing:
- 81 dilithium tests ✓
- 30 hdwallet tests ✓  
- 125 threshold lib tests ✓
- 9 derivation integration tests ✓
- 19 signing integration tests ✓
- 26 resharing tests ✓
- 16 signer state tests ✓

## References

- [Mithril Threshold Signatures](https://iohk.io/en/research/library/papers/mithril-stake-based-threshold-multisignatures/)
- [FIPS 204: ML-DSA Standard](https://csrc.nist.gov/pubs/fips/204/final)

---

> [!NOTE]
> **High Risk**
> New threshold cryptography (DKG, signing, resharing) plus dilithium API/encapsulation changes affect secret handling and compatibility; mistakes here are security-critical despite added zeroization and deserialization limits.
> 
> **Overview**
> Adds **`qp-rusty-crystals-threshold`** to the workspace: threshold ML-DSA-87 for NEAR MPC with **(t, n)** configs (validated up to **n ≤ 6** in this diff), **dealer** and **4-round Mithril DKG** keygen, multi-round signing with **SSID-bound** `borsh` broadcasts and **bounded deserialization**, **hyperball/K-iteration** tuning, **NEAR-style poke/message** drivers, **HD-style DKG contribution** derivation, benchmarks, and a hyperball parameter script.
> 
> **Dilithium** changes support that work: **`KeccakState`** gets **`Zeroize`/`ZeroizeOnDrop`** and drops **`Copy`**; **`SecretKey.bytes`** is private (callers use **`to_bytes()`**); **`fips202`**, **`packing`**, **`poly`**, **`polyvec`** are **`pub`** for reuse; **`Poly`** types also **`Zeroize`**. **hdwallet** tests switch to **`secret.to_bytes()`**.
> 
> Lockfile/workspace picks up **`borsh`**, **`libm`**, **`log`**, and related deps for the new crate.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50ae28acfe83e2f3fe27e4ab9cdeb9a7627ecaf8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13:
GPT 5.5 review

Reviewed [PR #74](https://github.com/Quantus-Network/qp-rusty-crystals/pull/74). I did not post anything on GitHub.

## Findings
- **High:** `recover_share` can silently use the wrong share in the `t == n` case. Unknown active parties are ignored, and if the computed mask is missing it falls back to the first map entry instead of failing.

```154:166:threshold/src/protocol/secret_sharing.rs
let active_key: u16 = active_parties.iter().fold(0u16, |acc, &p| {
	if let Some(idx) = dkg_participants.index_of(p) {
		acc | (1 << idx)
	} else {
		acc
	}
});

let share = shares.get(&active_key).or_else(|| {
	shares.values().next()
});
```

- **High:** normal share recovery silently skips missing translated subset shares, then returns a partially reconstructed secret. That should be an explicit error.

```249:283:threshold/src/protocol/secret_sharing.rs
if let Some(share) = shares.get(&u_translated) {
	// Convert share to NTT domain and accumulate
	// ...
}
```

- **High:** `ThresholdConfig` accepts all `n = 7` configs, but signing parameters do not define any `n = 7` entries and fall back to `2-of-2` values. That makes advertised `(2..7)-of-7` signing use the wrong rejection/hyperball parameters.

```80:85:threshold/src/config.rs
(2, 7) => 7,
(3, 7) => 55,
(4, 7) => 160,
(5, 7) => 320,
(6, 7) => 270,
(7, 7) => 650,
```

```366:382:threshold/src/protocol/signing.rs
match (config.threshold(), config.total_parties()) {
	// ...
	(6, 6) => (424124.0, 424197.0, 7.0),
	_ => (503119.0, 503192.0, 7.0), // Default to 2-of-2 values
}
```

- **High:** followers accept `Round4Retry` in any state. A premature or replayed retry can reset an honest follower mid-protocol and desynchronize the session.

```1102:1115:threshold/src/signing_protocol.rs
SigningMessage::Round4Retry => {
	if !self.is_leader() {
		self.retry_count += 1;
		// ...
		self.reset_for_retry();
	}
},
```

- **Medium:** `combine_with_message` accepts a fresh `message`/`context` instead of checking the values captured in `AfterRound3`. A caller can combine responses for one message while hashing another, producing silent invalid signatures.

```524:563:threshold/src/signer.rs
let (round2_data, my_responses) = match &self.state {
	SignerState::AfterRound3 { round2_data, my_responses, .. } =>
		(round2_data, my_responses),
	// ...
};

let signature_bytes = combine_signature(
	&self.public_key,
	&self.config,
	message,
	context,
	&w_aggregated,
	&all_responses,
)?;
```

- **Medium:** truncated Round 3 responses are padded with zero polynomials instead of rejected. This should fail early with `InvalidResponseSize`.

```621:637:threshold/src/protocol/signing.rs
if end <= data.len() {
	// unpack response
	responses.push(z);
} else {
	responses.push(polyvec::Polyvecl::default());
}
```

## Test Notes
`cargo test -p qp-rusty-crystals-dilithium` passed. `cargo check -p qp-rusty-crystals-threshold --no-default-features` passed with warnings.

`cargo test -p qp-rusty-crystals-threshold --lib --tests` did not finish: the main unit tests and derivation tests passed, but `test_threshold_matrix` and `test_threshold_matrix_dkg` were still running after more than 4 minutes, so I stopped the run. That runtime is worth fixing or gating, especially for CI.

n13:
Here's the comprehensive review of PR #74. I've organized findings by severity. The review covers cryptographic soundness and correctness across the new `threshold` crate and the `dilithium` security improvements.

---

# Review of PR #74: Threshold ML-DSA-87 + DKG + Resharing

## Overall Assessment

The PR is large (~16k LOC) and demonstrates a serious engineering effort. The signing protocol has correct rushing-attack defenses (HQ2), good zeroize hygiene, and 125 passing tests. **However, two critical cryptographic vulnerabilities make the resharing and key-derivation features unsafe to deploy as-is.** The README's own warning ("research/educational use only, not audited") is well-justified.

---

## Critical Findings (must fix before any production use)

### C1. Resharing Round 1 leaks individual shares in plaintext
**File:** `threshold/src/resharing/types.rs:343-358`, `threshold/src/resharing/protocol.rs:452-459`

`ResharingRound1Broadcast` simultaneously broadcasts both:
- `blinded_s1_contribution = recovered_share + blinding`
- `blinding_s1` (the blinding value itself)

Any passive observer trivially recovers `recovered_share = blinded - blinding`. Summing recovered shares from any `t` old-committee members reconstructs the full secret `s`.

The doc comment at `types.rs:336` confidently asserts this is safe:

```337:340:threshold/src/resharing/types.rs
/// **Security note:** Broadcasting the blinding values is safe because:
/// - The blinding alone doesn't reveal the secret
/// - The blinded contribution hides the actual share values
/// - Only the combination of all participants' shares (which requires threshold cooperation) can
///   recover the secret
```

This reasoning is wrong. The blinding doesn't hide anything when the value used for blinding is broadcast alongside the blinded value. This contradicts the README's claim that "Secret `s` never exposed during resharing."

### C2. Designated dealer reconstructs the entire secret in plaintext
**File:** `threshold/src/resharing/protocol.rs:1083-1107`

```1083:1097:threshold/src/resharing/protocol.rs
		// Compute the actual secret: blinded_total - total_blinding
		// Reduce modulo Q to ensure coefficients are in valid range
		let mut secret_s1 = vec![[0i32; N]; L];
		let mut secret_s2 = vec![[0i32; N]; K];

		for (secret_poly, (blinded_poly, blind_poly)) in
			secret_s1.iter_mut().zip(blinded_s1_total.iter().zip(total_blinding_s1.iter()))
		{
			for (secret_coeff, (blinded_coeff, blind_coeff)) in
				secret_poly.iter_mut().zip(blinded_poly.iter().zip(blind_poly.iter()))
			{
				let diff = blinded_coeff.wrapping_sub(*blind_coeff);
				*secret_coeff = reduce_coeff_mod_q(diff);
			}
		}
```

The "designated dealer" (smallest party ID in old committee) literally reconstructs `secret_s1` and `secret_s2` in memory, breaking the entire threshold property — anyone compromising the dealer during resharing gets the full key. Also, `secret_s1` / `secret_s2` are never zeroized (no `Zeroize` impl, no manual scrub), so they linger on the stack/heap after `deal_new_shares` returns.

This is a fundamentally different protocol from a proper proactive-resharing scheme (e.g., re-dealing of subset shares or polynomial-based resharing). It needs to be redesigned, not patched.

### C3. Key derivation uses publicly-derivable "secret material"
**Files:** `threshold/src/derivation.rs:70-92`, `threshold/src/keygen/dkg/protocol.rs:878-886`

`derive_dkg_contribution` uses `master_share.key()` as the per-party secret input:

```70:92:threshold/src/derivation.rs
pub fn derive_dkg_contribution(master_share: &PrivateKeyShare, tweak: &[u8; 32]) -> [u8; 32] {
	// Get secret material from the master share
	// We use the key field which contains the private key seed
	let secret_material = master_share.key();

	// Include party_id to ensure different parties get different contributions
	// even if they somehow had the same key material
	let party_id_bytes = master_share.party_id().to_le_bytes();

	// Use SHAKE256 for key derivation with domain separation
	// Input: domain || secret || party_id || tweak
	let mut state = fips202::KeccakState::default();
	fips202::shake256_absorb(&mut state, DKG_CONTRIBUTION_DOMAIN, DKG_CONTRIBUTION_DOMAIN.len());
	fips202::shake256_absorb(&mut state, secret_material, secret_material.len());
	fips202::shake256_absorb(&mut state, &party_id_bytes, party_id_bytes.len());
	fips202::shake256_absorb(&mut state, tweak, tweak.len());
	fips202::shake256_finalize(&mut state);

	let mut contribution = [0u8; 32];
	fips202::shake256_squeeze(&mut contribution, 32, &mut state);

	contribution
}
```

But for DKG-generated shares, `key()` is computed from purely public data:

```878:886:threshold/src/keygen/dkg/protocol.rs
	// Generate a deterministic key for this party
	let mut party_key = [0u8; 32];
	{
		let mut h = fips202::KeccakState::default();
		fips202::shake256_absorb(&mut h, &state.rho, 32);
		fips202::shake256_absorb(&mut h, &state.config.my_party_id.to_le_bytes(), 4);
		fips202::shake256_finalize(&mut h);
		fips202::shake256_squeeze(&mut party_key, 32, &mut h);
	}
```

`rho` is part of the public key bytes (`PublicKey::as_bytes()` exposes it via the standard ML-DSA-87 packing) and `party_id` is public. So `master_share.key()` is a deterministic function of public data, meaning **every party's "DKG contribution" for any tweak is publicly computable** by an external observer who only knows the master public key. The derived keys inherit no secrecy beyond what the tweak alone provides — and the tweak is public by design. The trusted-dealer path generates `key` from real entropy, but the DKG path makes it public.

The doc comment at `derivation.rs:39-44` claims "Only parties with valid master shares can compute their contribution" — this is false for the DKG path.

---

## High-Severity Findings

### H1. `unpack_responses` silently zero-pads malformed input
**File:** `threshold/src/protocol/signing.rs:613-641`

```613:641:threshold/src/protocol/signing.rs
pub(crate) fn unpack_responses(
	data: &[u8],
	config: &ThresholdConfig,
) -> ThresholdResult<Vec<polyvec::Polyvecl>> {
	let k = config.k_iterations() as usize;
	let single_response_size = L * 640; // L * POLY_LE_GAMMA1_SIZE
	let mut responses = Vec::with_capacity(k);

	for i in 0..k {
		let start = i * single_response_size;
		let end = start + single_response_size;

		if end <= data.len() {
			let mut z = polyvec::Polyvecl::default();
			for j in 0..L {
				let poly_start = start + j * 640;
				let poly_end = poly_start + 640;
				if poly_end <= data.len() {
					poly::z_unpack(&mut z.vec[j], &data[poly_start..poly_end]);
				}
			}
			responses.push(z);
		} else {
			responses.push(polyvec::Polyvecl::default());
		}
	}

	Ok(responses)
}
```

Truncated/short payloads from a malicious party silently produce all-zero `Polyvecl`s instead of erroring. Result: aggregations are corrupted (or signatures pass with bogus zero contribution from one party). This violates your "Fail early — never have a silent failure" rule. It should `Err(ThresholdError::InvalidResponseSize { expected, actual })` instead of padding with `Polyvecl::default()`.

### H2. `recover_share` (t == n) silently substitutes the wrong share
**File:** `threshold/src/protocol/secret_sharing.rs:162-167`

```162:167:threshold/src/protocol/secret_sharing.rs
		// Try to find the share with the active key first
		let share = shares.get(&active_key).or_else(|| {
			// Fallback: use any available share (like Go does with map iteration)
			shares.values().next()
		});
```

When the share for the exact `active_key` is missing, the code happily returns *any* share from the map. The "like Go does" comment is not a justification. This silently produces an incorrect signing share, which then fails later (or worse, succeeds in producing an invalid signature). Should be `ok_or(ThresholdError::InvalidSignatureShare { ... })`.

### H3. `get_threshold_params` defaults to 2-of-2 hyperball geometry for unknown (t, n)
**File:** `threshold/src/protocol/signing.rs:380-383`

```378:383:threshold/src/protocol/signing.rs
		(5, 6) => (479692.0, 479819.0, 7.0),
		(6, 6) => (424124.0, 424197.0, 7.0),
		_ => (503119.0, 503192.0, 7.0), // Default to 2-of-2 values
	}
}
```

`ThresholdConfig` allows `n` up to 7 (and `MAX_PARTIES = 7` in `error.rs:271`), but no `(_, 7)` row exists in the match. Any `n=7` configuration silently uses the 2-of-2 hyperball radii, which is wrong for the rejection-sampling distribution and breaks zero-knowledge of `s` (the response distribution becomes leakier than the analysis assumes). This should be a hard error rather than a silent fallback. Either add the missing rows or `panic!`/`return Err(...)` for unknown configs.

### H4. Round 4 (final signature) accepted by followers without verification
**File:** `threshold/src/signing_protocol.rs` (Round4Complete handling)

Followers accept the leader's combined signature on faith. The doc concedes this is acceptable for trusted committees, but the README of the crate also lists "untrusted parties supported." The follower should at minimum call `crate::verify_signature(public_key, msg, ctx, &sig)` before transitioning to `Done`; this is a cheap one-shot verification and removes the leader trust assumption entirely. (`verify_signature` is the function exposed at `threshold/src/lib.rs:217`.)

### H5. Resharing R2 messages rely on an external secure channel that isn't enforced
**Files:** `threshold/src/resharing/types.rs:392-417`, `threshold/src/resharing/protocol.rs` (Round 2 send path)

`ResharingRound2Message` carries new `s1`/`s2` shares as plaintext to a specific recipient via `Action::SendPrivate`. There is no in-protocol authenticated-encryption layer; the README assumes the transport gives confidentiality. If callers wire this up over a network without AE, every new share is on the wire in clear. Combined with C1/C2, this is "defense in depth violated." Recommend documenting this loudly in README and ideally adding an explicit `EncryptedShare<Recipient>` wrapper so the type system enforces it.

---

## Medium-Severity Findings

### M1. DKG `complete()` non-members skip Partial PK verification
**File:** `threshold/src/keygen/dkg/protocol.rs` (Round 4 `complete()` stage)

When a party isn't in subset `S`, it accepts whatever Partial PK is broadcast for `S` without verifying. Members of `S` *do* verify (by recomputing their own contribution), so as long as ≥1 honest member exists in every subset, an inconsistency would be caught — but if a subset is fully malicious, non-members will accept a forged Partial PK. The aggregation `t = Σ t_S` then yields a public key that doesn't match what honest members hold shares for. Recommend non-members at least check that `transcript_signature` from a subset member endorses the same Partial PK.

### M2. DKG R1 private secret stored by `subset_mask` allows DoS overwrite
**File:** `threshold/src/keygen/dkg/protocol.rs` (`MithrilDkg::message` handling of `Round1Private`)

The receiver doesn't validate that the sender is the legitimate leader for `subset_mask` before inserting into the per-subset map. A malicious non-leader can race or overwrite the leader's private contribution. R4 transcript verification catches the inconsistency, but the protocol then aborts — a denial-of-service primitive. Index by `(leader, subset_mask)` and only accept the leader's message.

### M3. `combine` fragility: malicious party with oversized z aborts the whole signature
**File:** `threshold/src/signer.rs` (`combine` / `combine_with_message`)

Round 3 aggregation sums all `z` responses; if any party submits a `z` that fails the `||z||∞ < γ1 - β` bound, *all* `K` rejection-sampling iterations may fail and the signature aborts. There's no per-party z-norm check before aggregation, so a single byzantine party can DoS the protocol. Consider validating each party's `z` bound before summing, then excluding the offending party.

Also: `BTreeMap<ParticipantId, _>` keyed insertion in Round 3 means a duplicate Round 3 from the same `party_id` silently overwrites. Should `Err(DuplicateBroadcast)`.

### M4. `signing_protocol.rs` participant-set divergence when `n > t`
**File:** `threshold/src/signing_protocol.rs` (`drop_party` and active set tracking)

When `total_parties > threshold`, different honest parties may drop different unresponsive participants depending on local timeout behavior, ending up with non-identical "active signer" sets and thus diverging Round 1/2 aggregations. The integration tests use `n == t` so this doesn't trigger. For `n > t`, you need a leader-driven explicit "active set" announcement that becomes part of Round 1 signing transcript. Currently this is a latent correctness bug for the most useful threshold case.

### M5. `pack_signature`'s `Result<...>` swallowed in `combine_signature`
**File:** `threshold/src/protocol/signing.rs:843-845`

```843:845:threshold/src/protocol/signing.rs
			if let Ok(sig) = pack_signature(&challenge_full, &z_centered, &hint) {
				return Ok(sig);
			}
```

If `pack_signature` returns an `Err`, the iteration continues to the next `k`, which masks any actual bug in packing. Either propagate the error or replace with a `?`. Today `pack_signature` happens to always return `Ok`, but the silent-failure pattern violates the "fail early" rule.

---

## Low-Severity / Defensive Findings

### L1. `single_commitment_size = 8 * 736` is a magic number
**File:** `threshold/src/signer.rs` Round 3 (`round3_respond`)

```text
let single_commitment_size = 8 * 736; // K * POLY_Q_SIZE
```

Replace with `K * POLY_Q_SIZE` constants from `dilithium_params` so this can't drift.

### L2. `HyperballSampleVector::new(size)` requires even `size`
**File:** `threshold/src/protocol/primitives.rs`

The internal `step_by(2)` over `samples[i+1]` would panic / corrupt if `size` were odd. Today every caller uses `N * (L + K) = 3840` (even), so it's latent. Add `debug_assert!(size % 2 == 0)` and document the precondition.

### L3. `generate_subsets_of_size` overflows `u16` if `n == 16`
**File:** `threshold/src/protocol/secret_sharing.rs`

`let max_val: u16 = 1 << n;` overflows when `n == 16`. Capped at 7 by `MAX_PARTIES`, so unreachable today. Use `u32` internally to be safe in case `MAX_PARTIES` is ever raised.

### L4. README in `threshold/src/keygen/README.md` describes a 5-round DKG; code is 4-round
**File:** `threshold/src/keygen/README.md`

The README documents Rounds 1–5 with a separate session-ID round, but the actual `MithrilDkg` is 4-round (commit → reveal → PK commit → PK reveal+sign). Sync the docs.

### L5. `validate_threshold_params` allows `t == n` but `n=2 t=2` is the "default" hyperball case
**File:** `threshold/src/error.rs:277-303`

That's not a bug per se, but combined with H3 it means you can construct configurations where the wrong rejection-sampling parameters are used without any warning.

---

## Confirmed Fixes from Prior Audit Items (good!)

- **HQ2 — Rushing adversary defense (commit-reveal in signing):** correctly implemented in `verify_commitment_hash` (`threshold/src/protocol/signing.rs:136`) and called from `round3_respond` (`threshold/src/signer.rs:383`). Tests in `signer_state_tests.rs:52` cover the adversarial case.
- **HQ4 — `SecretKey.bytes` is `private`:** confirmed in `dilithium/src/ml_dsa_87.rs`.
- **HQ5 — Seed zeroized after use:** `signer.rs:240`.
- **HQ8 — `KeccakState`: `Zeroize`+`ZeroizeOnDrop`, no `Copy`:** confirmed in `dilithium/src/fips202.rs`.
- **SHAKE256 squeeze bug:** `keccak_squeeze` previously reset `idx` inside the multi-block loop, causing each block to overwrite the start of the output buffer. Fix moved `let mut idx = 0;` outside the loop. This was a real correctness bug — large outputs (e.g., `tr` derivation, `mu`, response unpacking with multiple blocks) would have produced wrong bytes. The fix is correct.
- **HQ7 — Party management/timeout in `signing_protocol.rs`:** `dropped_parties` set, `BelowThreshold` error, etc.

---

## Recommendations (prioritized)

1. **Block all production deployment** until C1, C2, C3 are redesigned — these are not fix-by-patch issues; the resharing protocol and the derivation protocol need reworking.
2. **For C1/C2:** Replace plaintext-blinding with a proper proactive-resharing scheme (e.g., share the secret as a fresh Shamir/RSS polynomial whose constant term equals the old secret, sent privately and verified via PVSS or Feldman commitments). Never reconstruct `s` in the clear on any single machine.
3. **For C3:** `PrivateKeyShare.key` must be true entropy chosen during DKG (e.g., the per-party `seed` contribution before mixing) and not a deterministic hash of `rho || party_id`. Alternatively, derivation should hash actual share material (e.g., a Pedersen-commitment opening of `s_i`) rather than `key`.
4. **For H1/H2:** Apply your "fail early" rule literally — return errors instead of zero-padding or arbitrary fallbacks.
5. **For H3:** Make `get_threshold_params` total: either include all `(t, n)` rows allowed by `MAX_PARTIES`, or hard-error on miss.
6. **For H4:** Have followers verify `Round4Complete` via `verify_signature` unconditionally; the cost is negligible vs. the risk reduction.
7. **For M3/M4:** Add per-party z-bound checks before aggregation, and have the leader broadcast an authenticated active-set in Round 1 for `n > t` cases.
8. **Documentation:** Update `resharing/README.md` and `keygen/README.md` to reflect actual behavior. Remove the false "Secret `s` never exposed" claim until C1/C2 are fixed.

---

## What looks solid

- Module organization, error taxonomy, and zeroize hygiene on `PrivateKeyShare` / `Round1Data` / `Round2Data`.
- Commit-reveal scheme in signing (HQ2) is properly enforced before aggregation.
- `ParticipantList` (arbitrary ID → index mapping) is sound.
- 125 lib tests pass cleanly; integration tests exercise 2-of-3, 3-of-5, and DKG+sign end-to-end.
- The SHAKE256 fix is real and important.

The signing core (when invoked with a properly generated key) appears cryptographically sound modulo the `(t, n)`-parameter and silent-padding caveats above. The keygen DKG also looks structurally OK aside from the `key` derivation issue. The two protocols that need significant rework are **resharing** and **key derivation**, in that order.

If you want, I can go deeper on any single finding — for instance, sketching a corrected resharing flow or auditing a specific module line-by-line.

n13:
Reviewed by **GPT 5.5 XH**.

## Executive Summary

I reviewed PR #74 as a cryptographic implementation/security audit, focusing on threshold ML-DSA signing, DKG, resharing, secret handling, randomness/session binding, serialization, and Rust-side failure behavior.

Overall assessment: **not production-ready as a cryptographic security boundary as-is**. I did not find a critical forgery path or unauthorized-signature path, and the PR has solid work around SSID binding, participant checks, message-size caps in the signing path, zeroization in several areas, and standard ML-DSA verification of final signatures. However, there are several high-risk issues that should be addressed before relying on this in production.

I ran:

```bash
cargo test -p qp-rusty-crystals-threshold
```

Result: passed. Two measurement/statistical tests were intentionally ignored by the suite.

## Critical

None found.

## High

### H1. Resharing “forward secrecy” claim is false / overstated

Refs: `threshold/src/resharing/types.rs` around the entropy/forward-secrecy docs, `threshold/src/resharing/protocol.rs` session-seed and sub-share derivation.

Round 2 publicly reveals all entropy, and the resharing sub-shares are deterministically derived from old shares plus the resulting public session seed. An attacker who records the transcript and later compromises old shares can recompute the resharing randomness and derive the new shares.

Recommendation: remove the forward-secrecy claim or redesign resharing so post-compromise security does not depend on public transcript material. At minimum, document the actual threat model precisely.

### H2. Signing seed/nonce reuse can repeat threshold ML-DSA nonce material across messages

Refs: `threshold/src/protocol/signing.rs` `compute_ssid` / `generate_round1`, `threshold/src/signing_protocol.rs` SSID construction.

`compute_ssid` binds key/config/participants/attempt nonce, but not message/context. Round 1 randomness is derived from the raw `round1_seed` without also binding the SSID or message. If a caller accidentally reuses `round1_seed` and attempt nonce for another message with the same key/config/participants, the same nonce material can be reused under a different challenge.

Recommendation: derive Round 1 randomness from `domain || ssid || party_id || seed`, and include message/context or a message-bound attempt identifier in the SSID/nonce derivation contract.

### H3. Private protocol payloads rely entirely on external authenticated encryption

Refs: `threshold/src/resharing/types.rs` `ResharingRound4Message`, `threshold/src/keygen/dkg/types.rs` `Round1Private`.

Resharing sub-shares and DKG `K_S` are serialized in plaintext. Security depends completely on `SendPrivate` transport confidentiality, authenticity, and integrity. If integrators send these messages over unauthenticated or unencrypted channels, threshold secrecy is broken.

Recommendation: make AEAD-protected private transport a hard API contract, or add in-protocol authenticated encryption keyed from participant authentication keys.

### H4. Local rejection sampling emits zero Round 3 shares

Refs: `threshold/src/protocol/signing.rs` rejection loop and `pack_responses`.

When `response_float.excess(r, nu)` fails for an iteration, the loop continues and leaves `zs[i]` at its default zero vector. That zero vector is packed and broadcast. Zero vectors pass the later z-norm check and are treated as if the party contributed a valid share, which breaks the threshold linearity expectation for that iteration and increases liveness/correctness failures.

Recommendation: abort the signing attempt on local rejection, or encode per-iteration rejection explicitly and reject combination unless all required parties contributed valid shares.

## Medium

### M1. Documented old-committee resharing accusations are not implemented

Refs: `threshold/src/resharing/types.rs` Round 5 docs, `threshold/src/resharing/protocol.rs` Round 5 generation/combining.

The docs describe old-subset members independently recomputing dealer outputs and filing accusations. `ResharingRound5Broadcast` has no accusation field, and old-only parties effectively broadcast success with empty maps.

Recommendation: implement accusations/cross-verification or remove the claimed security property from the protocol documentation.

### M2. `existing_share` validation is missing at resharing construction

Refs: `threshold/src/resharing/protocol.rs` `ResharingProtocol::new`.

The constructor does not enforce that old committee parties provide a matching existing share, that new-only parties do not provide one, or that share party ID / participant list / public key are consistent. Misconfiguration fails late or can produce invalid dealing.

Recommendation: validate these invariants in `ResharingProtocol::new` and fail early.

### M3. Deserialization lacks semantic bounds on large maps

Refs: `threshold/src/keys.rs` `PrivateKeyShare`, `threshold/src/resharing/types.rs` Round 3/4/5 map-heavy messages, `threshold/src/keygen/dkg/types.rs` DKG map-heavy messages.

Several structures derive `BorshDeserialize` over `BTreeMap`s containing large share/public-key objects. Top-level frame caps help, but malformed valid-size payloads can still force unnecessary allocations/work or carry unexpected subset shapes.

Recommendation: add custom bounded deserializers and validate map lengths, subset masks, and expected subset counts against `(t, n)`.

### M4. DKG private shares and sender identity are transport-trust assumptions

Refs: `threshold/src/keygen/dkg/protocol.rs` `message(from, data)` and private-message handling.

`message(from, data)` checks `from` and claimed `party_id`, but does not cryptographically authenticate the envelope. Similarly, DKG `K_S` confidentiality is entirely transport-provided.

Recommendation: document this as a hard integration precondition and consider signed/encrypted message envelopes.

### M5. Round 3 signing responses are not validated at ingest

Refs: `threshold/src/signing_protocol.rs` Round 3 message handling.

Round 3 responses are stored without checking exact `response.len() == k * L * POLYZ_PACKEDBYTES`; malformed wrong-length responses fail later during combine.

Recommendation: validate Round 3 response size when received, matching the earlier Round 2 validation style.

### M6. Constant-time side-channel posture is not hardened

Refs: `threshold/src/protocol/primitives.rs` hyperball sampling and norm checks.

Hyperball sampling uses `f64`, Box-Muller, `libm`, and branchy rejection/norm checks. This may match the research/reference approach, but it should not be presented as hardened against local timing/FP side channels.

Recommendation: document this explicitly or replace with audited constant-time integer sampling/checking if local side-channel resistance is required.

### M7. Secret-share temporaries lack zeroization in some paths

Refs: `threshold/src/protocol/secret_sharing.rs` `SecretShare`; `threshold/src/resharing/protocol.rs` protocol state fields.

Several secret-bearing temporaries or protocol-state fields are not `Zeroize` / `ZeroizeOnDrop`, including recovered `SecretShare` values and resharing entropy/sub-share buffers.

Recommendation: derive or implement zeroization for all secret-bearing protocol state and temporary share containers.

## Low

- DKG seed and `h_keygen` derivation omit SSID; mix session identity into derivations for defense in depth.
- `PartialPublicKey.subset_mask` is not checked against its map key; reject inconsistent metadata.
- SSID mismatches and sender mismatches are often silently ignored; this is safe for many replay cases but weak for observability and “fail early” behavior.
- `ThresholdSigner` direct APIs lack the message-size guard present in `DilithiumSignProtocol`.
- `decompose_polyveck` assumes canonical `[0, Q)` inputs; normalize or assert at boundaries.
- Z aggregation uses Rust `% Q` rather than canonical modular reduction; currently likely latent, but should use the same reduction helpers everywhere.
- More full cross-implementation vectors against the CIRCL/Go threshold ML-DSA reference are needed.

## Priority Recommendations

1. Fix or remove the resharing forward-secrecy claim.
2. Bind threshold signing randomness to session/message context and fail hard on reuse-sensitive paths.
3. Require or implement AEAD for all private DKG/resharing payloads.
4. Stop broadcasting zero Round 3 shares on rejection.
5. Add bounded, validating deserializers for secret shares and map-heavy protocol messages.
6. Validate resharing `existing_share` invariants at construction.
7. Add adversarial tests for SSID mismatch/replay, bad existing shares, split dealer delivery, DKG `K_S` equivocation, and local rejection behavior.

n13:
# Security Audit — PR #74 "Threshold Signatures - 2"

> **Reviewer:** Claude **Opus 4.8 Max** (automated cryptographic security review). Findings are advisory; please confirm before acting on key‑custody decisions.

**Branch:** `illuzen/threshold-2` (HEAD `ec582d5`) · **Base:** `master`
**Scope:** +26,006 / −127 across 46 files. New `qp-rusty-crystals-threshold` crate (threshold ML‑DSA‑87) + hardening of the `dilithium` crate.
**Method:** read‑only review in a throwaway worktree; full read of the signing core, key types, NTT, and DKG/resharing/adapter deep‑dives, plus a clean `cargo test` run (all threshold tests pass, exit 0).

---

## 1. Executive Summary

This PR adds a large, ambitious threshold ML‑DSA‑87 implementation (DKG, signing, resharing, HD derivation, NEAR‑MPC adapter) plus genuine, well‑targeted hardening of the base `dilithium` crate. The engineering quality is high in several respects: `#![deny(unsafe_code)]`, bounded `borsh` deserialization, SSID session‑binding, commit‑reveal with hash verification, public‑key/TR binding, follower‑side signature verification, and `Zeroize`/`ZeroizeOnDrop` on the principal key types. Functional correctness is backed by an extensive passing test suite.

However, **the construction is research‑grade and its safety is almost entirely contingent on an unenforced precondition: the caller must feed a unique, cryptographically‑random 32‑byte seed for every signing session.** The crate contains **no CSPRNG at all** — every "random" value (signing nonce, DKG contribution, resharing entropy, dealer keygen) is SHAKE256‑expanded from a caller‑supplied seed — and the masking nonce that hides the secret key is derived deterministically from that seed. The only signing entry point takes the raw seed, the SSID/`attempt_nonce` is *independent* of it (so freshening the nonce does **not** freshen the secret randomness), and the shipped examples use trivially predictable seeds. Seed reuse across two messages yields **full secret‑key recovery** (Fiat‑Shamir‑with‑aborts nonce reuse). The library's own `lib.rs` states it "has not undergone a security audit and should not be used in production," which is in tension with the PR's stated NEAR‑MPC production target.

Secondary concerns: the secret nonce is sampled with **floating‑point** transcendental functions (`log`/`sin`/`cos`/`sqrt`) and the secret‑share arithmetic is **not constant‑time**, both of which are side‑channel exposures that standard ML‑DSA was specifically designed to avoid; several adversarial conditions are swallowed as `Ok(())`; and resharing state is not zeroized.

**Overall assessment: NOT READY for production / high‑value key custody in its current form.** The cryptographic protocol logic appears internally consistent and is well‑tested for the honest path, but (a) the unsafe‑by‑default randomness API is a critical footgun, (b) side‑channel resistance is not met, and (c) the underlying threshold construction and its hardcoded security parameters have not been independently verified. Suitable for continued R&D; gate production use behind a safe‑by‑default randomness API, a constant‑time/side‑channel pass, and an external cryptographic review of the scheme + parameters.

| Severity | Count |
|---|---|
| Critical | 1 |
| High | 3 |
| Medium | 6 |
| Low | 5 |

---

## 2. Critical

### C‑1 — Unsafe‑by‑default randomness: deterministic nonce derivation enables full key recovery on seed reuse
**Files:** `threshold/src/signer.rs:302‑320`, `threshold/src/protocol/signing.rs:258‑309`, `threshold/src/signing_protocol.rs:548‑637, 789‑796`, `threshold/Cargo.toml:14‑15`

The secret masking nonce `y` (the lattice analogue of the ECDSA nonce `k`) is derived **deterministically** from a caller‑supplied seed:

```rust
// protocol/signing.rs — generate_round1
fips202::shake256_absorb(&mut state, seed);      // <-- ALL session randomness comes from `seed`
fips202::shake256_absorb(&mut state, b"rho_prime");
...
iter_seed.copy_from_slice(seed);                 // per-iteration nonces also derive only from `seed`
```

The only Round‑1 entry point is `round1_commit_with_seed(ssid, seed)`; there is **no RNG‑based variant**. A grep of the whole crate shows `rand`/`OsRng`/`getrandom` appear **only in doc comments and READMEs**, never in library code — yet `rand`/`rand_core` are declared dependencies (dead, misleading). The NEAR adapter `DilithiumSignProtocol::new(...)` likewise *takes* `round1_seed` as a parameter and never generates it.

Why this is catastrophic, not merely theoretical:
- Each party broadcasts `z_i = c·s1_i + y_i`. Two sessions over different messages with the **same seed** reuse `y_i` but use different challenges `c, c'`, so `z_i − z_i' = (c−c')·s1_i` — a linear equation in the secret share. A handful of such pairs recovers `s1_i`; t shares reconstruct the full signing key.
- **The footgun is unusually sharp:** `seed` (which controls `y`) and `attempt_nonce` (which controls the SSID, `protocol/signing.rs:90‑128`) are *independent*. An integrator who diligently rotates `attempt_nonce`/`ChannelId` per attempt but reuses `round1_seed` still leaks the key. The SSID does **not** include the message either.
- The shipped examples actively demonstrate the unsafe pattern: `lib.rs:40‑43` (`round1_seed[0] = i as u8`) and tests use `session_seed[0] = attempt`.

**Recommendation (highest priority):**
1. Provide a safe‑by‑default API: a `round1_commit(&mut self, rng: impl RngCore + CryptoRng)` (and an adapter constructor) that samples `round1_seed` internally from `OsRng`/`getrandom`; relegate `*_with_seed` to a clearly‑named `*_deterministic_for_testing` or feature‑gated function.
2. Bind the nonce derivation to the SSID (and ideally message) so that a fresh `attempt_nonce` *forces* fresh `y`, removing the independence footgun.
3. Remove the unused `rand`/`rand_core` deps or actually use them; delete the dead `RandomnessError` variant or wire it up.
4. Remove unsafe seeds from all doc examples; make the doctest examples use a CSPRNG.

---

## 3. High

### H‑1 — Floating‑point sampling of the secret nonce (side‑channel + portability risk)
**File:** `threshold/src/protocol/primitives.rs:284‑341, 346‑417`

`HyperballSampleVector::sample_hyperball` generates the secret nonce using Box‑Muller over `f64` with `libm::log`, `libm::sqrt`, `libm::cos`, `libm::sin`, and division, and the rejection decision `excess()` is FP arithmetic on secret‑dependent data:

```rust
let ln_f1 = libm::log(f1);
let sqrt_neg2ln = libm::sqrt(-2.0 * ln_f1);
let z1 = sqrt_neg2ln * libm::cos(2.0 * PI * f2);
```

ML‑DSA deliberately avoids floating point precisely because transcendental FP routines are **variable‑time** (data‑dependent), creating a timing side channel on the very value that must stay secret. (`libm` does give cross‑platform determinism, which is good for correctness, but does not make these ops constant‑time.) This is inherent to the hyperball approach but remains a genuine side‑channel exposure for an HSM/MPC‑node deployment.

**Recommendation:** Treat side‑channel resistance as a deployment requirement and document the FP exposure prominently; explore constant‑time/integer Gaussian or rejection alternatives, or restrict deployment to environments where wall‑clock timing of the sampling step is not observable to adversaries. Have the scheme's side‑channel posture reviewed by a lattice‑crypto specialist.

### H‑2 — Non‑constant‑time arithmetic on secret key shares
**Files:** `threshold/src/circl_ntt.rs:85‑99, 109‑146, 155‑199`; `threshold/src/protocol/primitives.rs:43‑66`; `threshold/src/protocol/signing.rs:586‑614`

The crate's own NTT and modular reduction operate on secret data with **data‑dependent branches** (sign tests), e.g. `mul_hat`/`ntt` convert coefficients via `if *a_coeff < 0 { ... }`, and `recover_share`/`generate_round3_response` run these over the secret shares `s1`/`s2` and over `c·s1`. While `mont_reduce_le2q` itself is branchless, the surrounding `if c < 0` / `if coeff > Q/2` reductions are not, leaking via branch timing.

**Recommendation:** Make secret‑dependent reductions branchless (mask‑select, as the base dilithium reduce/`le2q_mod_q` already do at `primitives.rs:43‑48`) across the NTT and response paths; add a constant‑time review checklist for any function that touches `s1`/`s2`/`y`.

### H‑3 — Non‑standard construction with hardcoded, in‑tree‑unverified security parameters
**Files:** `threshold/src/protocol/signing.rs:399‑420` (hyperball `r, r', nu`), `threshold/src/config.rs:62‑84` (`k_iterations`), `threshold/scripts/compute_hyperball_params.py`

Soundness of the masking (zero‑knowledge of the transcript, i.e. that signatures don't leak the key) depends entirely on the rejection radius pair `(r, r')` and `nu` used by `excess(r, nu)`. These are hardcoded floats "from the reference implementation," and the effective secret key here is the **sum of many η‑bounded RSS shares** (e.g. ≈[−40,40] for (4,6) vs. standard η=2), so the MLWE/rejection analysis differs from textbook ML‑DSA. The derivation script is a **Monte‑Carlo simulation** (statistical, not a proof) and there is no test asserting the Rust constants match it. If `r`/`r'` are too loose, the transcript leaks the key over many signatures; too tight, only availability suffers.

**Recommendation:** Obtain/publish the formal parameter derivation and a citation to the security proof for this exact RSS+hyperball construction; add a test that regenerates and checks the constants against the script; have the scheme reviewed externally before production. Reconcile `lib.rs:14` ("supports up to (7,7)") with the enforced `MAX_PARTIES = 6`.

---

## 4. Medium

### M‑1 — DKG subset secret `K_S` is not commitment‑bound; first‑message‑wins
**File:** `threshold/src/keygen/dkg/protocol.rs:606‑676, 952‑957`; `types.rs:384‑395`
Round‑1 commitments cover each party's broadcast randomness `r_i` but **not** the per‑subset secret `K_S`, which a subset leader sends in cleartext P2P and which is accepted first‑come (`or_insert`). Misbehavior is only caught by the Round‑4 algebraic PK check (abort‑only, no dispute/attribution protocol). A malicious/erroneous leader can grief to a late abort. **Recommendation:** commit to `K_S` in Round 1 (or add cross‑member `K_S` consistency + a complaint round); document that `SendPrivate` requires an authenticated‑encrypted channel.

### M‑2 — Resharing: documented dealer cross‑verification not implemented; missing `existing_share` validation
**File:** `threshold/src/resharing/protocol.rs:305‑330, 854‑875`; `types.rs:584‑588`
The README/types describe old‑committee members recomputing `r_{I→J}` and filing `DealerAccusation`s, but no such type or logic exists; only new‑committee members verify (commitment + final PK‑preservation at `protocol.rs:1245‑1252`). `ResharingProtocol::new` stores `existing_share` with **no** validation against the config/public key despite test comments claiming otherwise. Detection of a cheating dealer therefore reduces to an unattributed global abort. **Recommendation:** implement the documented cross‑verification/accusation path or remove the claim; validate `existing_share.party_id()/threshold()/dkg_participants()` and PK consistency in `new`.

### M‑3 — Adversarial inputs swallowed as `Ok(())` (violates fail‑loud)
**Files:** `signing_protocol.rs:1044‑1052`; `keygen/dkg/protocol.rs:540‑557`; `resharing/protocol.rs:463‑474`
SSID mismatch, sender/`party_id` mismatch, and non‑participant messages return `Ok(())` (some without even a `warn!`). The `SsidMismatch` error variant is never returned. This hides active attacks unless logs are monitored and can produce silent liveness stalls — contrary to the project's "no silent failures" rule. **Recommendation:** return typed errors (or a structured "rejected" signal) for these conditions so callers can detect/alert.

### M‑4 — Secret material not zeroized in resharing state and the signing adapter
**Files:** `resharing/protocol.rs:241‑293` (struct holds `seed`, `my_entropy`, `existing_share`, sub‑shares with no `Zeroize`/`Drop`); `signing_protocol.rs:493‑496` (`round1_seed` never zeroized; no `Drop`)
The core `ThresholdSigner`/`PrivateKeyShare`/`Round1Data` are correctly zeroized, but the resharing protocol object and the adapter's `round1_seed` (which *is* the key‑leaking secret from C‑1) are left in memory. **Recommendation:** implement `Zeroize`/`ZeroizeOnDrop` on `ResharingProtocol` and `DilithiumSignProtocol` (zeroize `round1_seed`, entropy, sub‑shares, `existing_share`).

### M‑5 — `Debug`/`Clone` on secret‑bearing message types (leak if logged)
**Files:** `keygen/dkg/types.rs:398‑406` (`Round2Broadcast` derives `Debug` over revealed `randomness`); `resharing/types.rs:537‑546` (`ResharingRound4Message` is `Debug + Clone` over `NewShareData` plaintext shares)
A debug/log of these structs would print secret material. **Recommendation:** remove `Debug` or implement redacting `Debug`; minimize cloning of share material.

### M‑6 — Secret‑key coefficient distribution deviates from standard ML‑DSA η bound
**File:** `threshold/src/keygen/dealer.rs:241‑305`; `keys.rs:252‑301` (coefficient‑growth monitors)
The reconstructed `s1/s2` are sums of `C(n, n−t+1)` independent η‑bounded shares, producing a wider secret than ML‑DSA's η=2. This is intrinsic to the RSS design and the hyperball params claim to compensate, but it changes the MLWE distribution and interacts with H‑3. The presence of `coefficient_stats`/`collect_all_coefficients` monitors suggests growth (esp. after repeated resharing) is a known concern, yet there is **no runtime norm bound enforced** before shares are used for signing. **Recommendation:** enforce an explicit norm/range check on shares post‑DKG and post‑resharing; tie acceptance to the same bound assumed by the hyperball analysis.

---

## 5. Low

- **L‑1 — Dead/misleading randomness deps.** `rand`/`rand_core` are declared (`Cargo.toml:14‑15`) but unused in library code; `ThresholdError::RandomnessError` (`error.rs:77`) is never produced. Remove or use; otherwise it implies an RNG that isn't there.
- **L‑2 — Public utility can panic/wrap on shift overflow.** `generate_subsets_of_size` (exported via `lib.rs:190`) computes `1u16 << n` (`secret_sharing.rs:105`); `n ≥ 16` panics (debug) or wraps (release). Validate `n` or use checked shifts.
- **L‑3 — Round‑4 signature lacks exact‑size bound on deserialize.** `Round4Broadcast.signature: Vec<u8>` (`signing_protocol.rs:237‑242`) is bounded only by the 4 MB outer cap, unlike the `SIGNATURE_SIZE`‑exact `Signature` (`broadcast.rs:240‑259`); wrong‑length sigs are dropped silently (liveness, not memory). Add the exact‑size check.
- **L‑4 — Doc/impl mismatches.** `signing_protocol.rs:22‑23` claims "does not include cryptographic session identifiers" (outdated — SSID is enforced); `primitives.rs:506‑514` `poly_pack_w` docs claim a debug panic on coeff ≥ Q that isn't implemented; `lib.rs:14` "(7,7)" vs `MAX_PARTIES=6`. Misleading security docs are a hazard in crypto code.
- **L‑5 — `verify_bounds()` defined but never called** for DKG contributions (`keygen/dkg/types.rs:323‑340`), so bound safety relies solely on `uniform_eta` correctness. Either call it or remove it.
- **Informational — widened dilithium API surface.** `fips202`, `packing`, `poly`, `polyvec` are now `pub` (`dilithium/src/lib.rs:91‑96`), exposing low‑level, non‑constant‑time primitives to all consumers. Needed for crate reuse, but document the sharp edges.

---

## 6. Dilithium crate changes (reviewed — net positive)

- `KeccakState`: drops `Copy`, adds `Zeroize`/`ZeroizeOnDrop` (`fips202.rs:16`). Good — prevents silent copies of sensitive sponge state.
- `SecretKey.bytes` made private with `to_bytes()`/`from_bytes()` (`ml_dsa_87.rs:133‑156`); `hdwallet` tests updated accordingly. Good encapsulation; no behavior change.
- `Poly`/`Polyveck`/`Polyvecl` add `Zeroize` (`poly.rs`, `polyvec.rs`). Good.
- The PR description's "SHAKE256 `keccak_squeeze` multi‑block fix" is **already in `master`** (via #78); it's not part of this diff. The current `keccak_squeeze` (`fips202.rs:352‑371`) is correct for multi‑block squeezes.

---

## 7. What's done well

`#![deny(unsafe_code)]`; bounded `borsh` deserialization with pre‑allocation size checks and tests (`broadcast.rs`, `participants.rs:86‑119`); SSID session‑binding + commit‑reveal with hash verification before aggregation (`protocol/signing.rs:163‑190`, `signer.rs:457‑497`); public‑key/TR binding to reject poisoned keys (`signer.rs:213‑221`, `keys.rs:53‑66`); followers verify the leader's signature against the public key (`signing_protocol.rs:962‑976`); DKG transcript‑signature consensus and resharing PK‑preservation byte‑equality check; `ParticipantList`/`PrivateKeyShare` zeroization and redacted `Debug`; comprehensive, passing test suite (verified locally).

---

## 8. Prioritized recommendations

1. **Fix C‑1 first:** add a CSPRNG‑backed, safe‑by‑default signing/keygen API; bind nonce derivation to the SSID; demote `*_with_seed` to test‑only; purge unsafe seeds from examples.
2. **Side‑channels (H‑1/H‑2):** branchless secret‑dependent reductions; document/limit the FP sampling exposure; get a lattice‑side‑channel review.
3. **Validate the scheme & parameters (H‑3, M‑6):** external review of the RSS+hyperball construction; in‑tree test pinning `(r, r', K)` to the derivation script; enforce share norm bounds.
4. **Fail loud (M‑3)** and **zeroize all secret state (M‑4)**; remove `Debug`/reduce clones on secret messages (M‑5).
5. **Resharing robustness (M‑2)** and **DKG `K_S` binding (M‑1):** implement (or remove claims about) dealer cross‑verification; validate `existing_share`.
6. Address the Low items and doc mismatches.
7. Until 1–3 are complete, keep the `lib.rs` "research only — not for production" warning authoritative and do not custody high‑value keys with this crate.

The PR is a strong R&D milestone, but I'd recommend **not merging it as production‑ready**: at minimum land C‑1 and the fail‑loud/zeroize fixes in this PR, and schedule H‑1/H‑2/H‑3 + an external cryptographic review before any production NEAR‑MPC use.

---

<sub>Automated security review by Claude Opus 4.8 Max. Line numbers reference branch HEAD `ec582d5`.</sub>

n13:
**Supplemental review comments — Grok build 0.1**

Building on the prior review (by 5.5) and the Cursor Bugbot automation already present on the thread, here are only the additional / refined observations from a fresh pass on a clean throwaway worktree (current PR head). I did not re-cover ground already addressed.

### Additional / refined findings

- **Floating-point (f64 + libm) in the core hyperball sampler for threshold masking/rejection.**  
  `HyperballSampleVector` stores a `Box<[f64]>` and `sample_hyperball` (plus `round_z_response`) performs the entire mask generation for the K parallel attempts via SHAKE256 → u64→f64 scaling → Box-Muller (libm `log`/`sqrt`/`cos`/`sin` + `PI`) → radius scaling → later `libm::round`.  
  This path is security-critical: the samples are what keep the *combined* (t-of-n) mask inside the rejection-sampling bounds that the precomputed K values and (r, r') hyperball parameters were tuned for.  
  Risks not previously highlighted: (a) libm transcendental functions are typically variable-time and table-driven (potential side-channel on many targets); (b) cross-platform FP reproducibility/rounding/denormal behavior can affect whether a given attempt succeeds or is rejected, potentially turning statistical tuning assumptions into observable bias or liveness differences; (c) the sampler is fed from attempt-specific `rhop` derived from the caller's `round1_seed`/`session_seed`.  
  File: `threshold/src/protocol/primitives.rs:270` (struct + Zeroize), `284` (sample_hyperball), `304` (`try_into().unwrap()` on SHAKE output slices inside the loop), `316-321` (Box-Muller with libm), `337` (final sqrt factor), `350` (round + reduction).  
  The K table and python script assume a specific distribution; any divergence here affects the "K iterations are enough" claim for the larger configs (e.g. 5-of-6, K=380).

- **`.unwrap()` / `.expect()` on "impossible" cases still present in non-test library code.**  
  Examples include dealer keygen assuming sequential DKG indices will always be valid (`.expect("sequential IDs are always valid")`) and several setup paths in keys/derivation that use `unwrap` even outside pure `#[cfg(test)]` blocks. Combined with the ResharingConfig gap already noted by automation (no `n <= 6` / hyperball guard, later `.expect`), a malformed but well-formed-looking committee or DKG participant list supplied over the wire (or during handoff) can still produce a panic instead of a clean `ThresholdError` / `SignProtocolError`.  
  Locations: `threshold/src/keygen/dealer.rs:185`, `threshold/src/keygen/dealer.rs:337` (and the ResharingConfig case previously flagged), plus setup sites in `derivation.rs` and `keys.rs` that leak into lib surface under test helpers.  
  Recommendation: drive everything through the existing rich error types (or a small internal `bail!` macro) so there are no panics on authenticated-but-invalid protocol messages.

- **Dealer keygen NTT mix (corroboration + test gap).**  
  The automation correctly called out that `generate_with_dealer` produces `s1h_total` via `circl_ntt` but then builds the public key `t` using the dilithium crate's `pointwise_montgomery` + `invntt_tomont` (lines ~117-125). Signing and DKG paths are circl_ntt-only. While the two NTTs are mathematically equivalent today, the dealer tests only check that keygen "succeeds" and that shares round-trip; they do not perform an end-to-end threshold signature with the resulting dealer keys under the real (circl_ntt) signing path and verify it against the packed public key. A drift between the two NTT implementations would produce keys that appear to generate but later fail (or produce invalid signatures) only at signing time.  
  Locations: `threshold/src/keygen/dealer.rs:116-125` (t computation) and the `generate_threshold_shares` call just above; cross-ref the circl_ntt conversion safety note already flagged by automation.

- **Weak iteration domain separation for K > 255 is worse than a single-byte collision in the seed.**  
  The automation noted the `k_iter as u8` XOR into `iter_seed[0]` (and high byte into [31]) plus the separate absorb of `k_iter as u8`. For K=380 this means many pairs of iterations have nearly identical domain separation. In addition, the hyperball sample itself (see first item) is derived from the same `rhop` + nonce material; any weakness in the iter seed directly affects the statistical independence of the parallel attempts that the whole K-tuning rests on.  
  Location: `threshold/src/protocol/signing.rs:208-212` (and the call sites that feed it into `sample_hyperball`).

- **HyperballSampleVector holds secret-derived mask material in f64 form; its Zeroize is present but the surrounding usage (BTreeMap of responses, aggregation, rounding) should be audited for any intermediate copies that escape the zeroize scope before the final signature is emitted.** The struct itself is good (`Zeroize + ZeroizeOnDrop`), but the values are the actual "randomness" that hides the threshold s1/s2 shares.

These are the deltas I have that were not already surfaced in the 5.5 review or the existing automation comments. Everything else I looked at (zeroization hygiene on the new `PrivateKeyShare`/`SignerState`/`SecretShareData` paths, SSID binding, exact-t enforcement, TR poisoning defense, bounded deserialization, NTT accumulator u64 overflow protection, etc.) was already well covered or matches the positive notes in the prior review.

**Grok build 0.1** (cryptographic security auditor pass on throwaway worktree `/tmp/qp-rusty-crystals-pr-74-supplement-1780723424`). Happy to drill into any of the above (especially the FP sampler vs. reference or full DKG/resharing transcript consistency) if useful.

n13:
## Claude Opus 4.8 re-review (issues addressed)

Re-reviewed at head `d075a749` in a throwaway worktree. Ran the threshold lib suite: **201 passed, 0 failed**. Nothing was changed on the branch by this review.

The prior review rounds (GPT 5.5, GPT 5.5 XH, Claude Opus 4.8 Max, Grok) have been worked through. Almost every concrete, code-level finding is now fixed, and the two structural Critical findings (resharing + key derivation) have been genuinely redesigned rather than patched.

### Critical — all resolved
- **Resharing leaked shares in plaintext (blinding + blinded value broadcast together) / designated dealer reconstructed the whole secret.** Rewritten as distributed per-subset re-sharing: each dealer only re-shares its own `s_I^old` (which it already holds) into sub-shares `r_{I→J}`; Rounds 1–3 broadcast only hash commitments; Round 4 sub-shares travel over a required authenticated-encrypted channel. No path reconstructs `s` anymore. Cheating-dealer detection is real: sub-shares checked against Round 3 commitments, new-subset members cross-verify identical `s_J^new`, and the original public key is reconstructed byte-for-byte from `Σ t_J` (`verify_public_key_preservation`).
- **Key derivation used publicly-derivable material** (`master_share.key()` = `H(rho‖party_id)` in the DKG path). `derive_dkg_contribution` now hashes the actual secret `s1`/`s2` polynomials (`hash_secret_shares`), with a regression test (`test_contribution_independent_of_legacy_key_field`).
- **Deterministic-nonce / seed-reuse footgun.** Round 1 nonce derivation now binds the SSID (`seed ‖ ssid ‖ "rho_prime"`); unused `rand`/`rand_core` library deps removed; dangerous `round1_seed[0] = i` examples replaced with `rand::random()` + explicit "reusing seeds enables full key recovery" warnings.

### High — resolved
- `recover_share`: the "use any share" / "ignore unknown active party" fallbacks are gone — both now hard-error, with regression tests.
- `unpack_responses`: no longer zero-pads; hard-errors `InvalidResponseSize` on length mismatch.
- Threshold/hyperball params: `get_hyperball_params` returns `Option` with `ok_or_else` at all call sites (no silent 2-of-2 fallback); `MAX_PARTIES = 6`, and `n=7`/unsupported configs rejected at construction.
- `Round4Retry` removed entirely (combination failure is terminal); followers now `verify_signature` on the leader's output before accepting (`test_follower_rejects_invalid_signature_from_leader`).
- Zero Round-3 responses excluded from combination + per-party z-norm checks; duplicate Round-3 broadcasts error (`DuplicateBroadcast`).
- Forward-secrecy claim removed / downgraded to "session randomization (not post-compromise)"; secure-channel requirement for private payloads is now a documented hard contract.

### Medium — resolved / mitigated
- Resharing coefficient growth: `SUBSHARE_COEFF_BOUND` enforced on every received sub-share.
- DKG `K_S`: Round1Private rejected from non-leaders and non-member recipients (fixes DoS-overwrite); residual first-message-wins documented as relying on the Round 4 PK-abort check.
- Bounded borsh deserialization with malicious-length-prefix tests; unused error variants removed.
- Fail-loud: ignorable messages (SSID/sender/non-participant) now emit `warn!` before `Ok(())`.
- Slow `test_threshold_matrix` split per-`n`; longest resharing tests `#[ignore]`-gated.

### Residual / minor (non-blocking)
1. `combine_with_message` still uses the caller-supplied `message`/`context` and ignores the values captured in `AfterRound3` (the safe `combine()` and the network adapter use the captured ones, so this is a footgun, not an exploit — a wrong message yields `CombinationFailed`, not a forgery).
2. `lib.rs` doc still says "supports configurations up to (7, 7)" while `MAX_PARTIES = 6`.
3. By-design / not closable here (correctly handled via the retained "research only — not audited" warning): no in-library CSPRNG (caller supplies the seed) and the SSID does not bind the message; floating-point/`libm` hyperball sampling and non-constant-time NTT on secret data (now documented); hardcoded hyperball/`K` parameters still rest on a Monte-Carlo script rather than a published proof — these remain the items needing an external cryptographic review before any production use.

**Verdict:** The issues from the prior reviews have been addressed. The three Critical items are properly fixed (two via real redesigns), the silent-failure/fallback Highs are now hard errors with tests, and what remains is one low-severity API footgun, a stale doc line, and the inherent "research-grade, needs external audit" posture the crate already declares. A small follow-up PR addresses items 1 and 2.

n13 (approved):
GTG

PR #75, opened 2026-05-03
Fix issues in threshold 2

n13:
Fixes on top of https://github.com/Quantus-Network/qp-rusty-crystals/pull/74 

All done! Here's the summary of what was fixed.

## Summary

**Test results — all green:**
- 128 lib tests · 26 resharing tests · 9 derivation tests · 16 signer-state tests · 19 integration tests = **198 passing, 0 failing**

### C3 — derived keys no longer publicly recoverable

In `threshold/src/derivation.rs`:
- `derive_dkg_contribution` now hashes the actual secret share polynomials (`s1`, `s2`) instead of the `key` field (which the DKG path derived from the public `rho || party_id` and was therefore publicly recoverable). Domain separator bumped to `v2`.
- Added `hash_secret_shares` helper that canonically serializes every `(subset_mask, s1, s2)` triple from the share's `BTreeMap`.
- Added `test_contribution_independent_of_legacy_key_field` — two shares that differ only in the publicly-derivable `key` byte string must now produce the same contribution.

In `threshold/src/keygen/dkg/protocol.rs` and `threshold/src/resharing/protocol.rs`:
- The `party_key` field is now derived from real entropy (the actual share polynomials) rather than a hash of public values. This is hygiene; derivation no longer depends on it.

### C1 + C2 — resharing redesigned around per-subset distributed re-sharing

**The core insight:** for each old RSS subset `I`, the lowest-ID member `D_I` deterministically derives sub-shares `r_{I→J}` for every new subset `J` such that `Σ_J r_{I→J} = s_I^old`. New shares are then `s_J^new = Σ_I r_{I→J}`, which preserves `Σ_J s_J^new = s` without anyone ever holding `s` in clear.

In `threshold/src/resharing/types.rs`:
- Removed `BlindedContribution`, `blinded_s1_contribution`, `blinded_s2_contribution`, `blinding_s1`, `blinding_s2`, `blinding_commitment` from `ResharingRound1Broadcast` (these were the C1 leak — broadcasting both blinded share and the blinding value lets anyone subtract).
- `ResharingRound1Broadcast` now carries only `commitments: BTreeMap<SubsetPair, [u8;32]>` (hash commitments, no plaintext).
- `ResharingRound2Message` now carries `contributions: BTreeMap<SubsetPair, NewShareData>` keyed by `(old_subset, new_subset)` for private peer-to-peer delivery.
- `ResharingRound3Broadcast` gained an `accusations: Vec<DealerAccusation>` field for cross-verification.

In `threshold/src/resharing/protocol.rs` (full rewrite):
- Round 1: each old participant who is the designated dealer for any subset broadcasts hash commitments.
- Round 2: each dealer privately delivers the sub-shares each recipient needs.
- Round 3: new-committee parties verify received sub-shares against commitments, sum them into `s_J^new`, and commit; old-committee parties independently recompute and accuse misbehaving dealers.
- `derive_subshares` PRF-derives sub-shares from `SHAKE256(s1_I || s2_I)`, with the residual sub-share absorbing the sum constraint. **No party ever computes `s = Σ_I s_I^old`** — the C2 single-point-of-compromise is gone.
- Subtle bug caught during testing: `old_subset_order` had to be a *canonical* enumeration (same for every party), not derived from each party's individual share keys — otherwise dealers and verifiers compute different `residual_idx` values and accusations explode.

In `threshold/src/resharing/{mod.rs, README.md}`:
- Updated documentation to describe the new protocol accurately and dropped the (false) "Secret never exposed" claim of the old design.

### What's next
The remaining findings from the original review (H1–H5, M1–M5, L1–L3) are still untouched — happy to tackle any of them whenever you're ready. The most impactful next ones are probably H1 (silent zero-fill in `unpack_responses`) and H2 (silent fallback in `recover_share` for `t=n`).

### DIFF
--- a/threshold/src/derivation.rs
+++ b/threshold/src/derivation.rs
@@ -21,16 +21,20 @@
 //!
 //! # Security
 //!
-//! The DKG contribution uses secret material from the master share, ensuring
-//! that outsiders cannot compute the derived shares even though the tweak
-//! is public.
+//! The contribution input is bound to the share's actual secret polynomials
+//! (`s1`, `s2`), which are the only true secret material in a `PrivateKeyShare`.
+//! The legacy `key` byte string is *not* used here, because in the DKG path it is
+//! derived deterministically from public values (`rho`, `party_id`) and therefore
+//! provides no secrecy.
+
+use alloc::vec::Vec;
 
 use qp_rusty_crystals_dilithium::fips202;
 
 use crate::keys::PrivateKeyShare;
 
 /// Domain separator for DKG contribution derivation.
-const DKG_CONTRIBUTION_DOMAIN: &[u8] = b"near-mpc-dilithium-dkg-contribution-v1";
+const DKG_CONTRIBUTION_DOMAIN: &[u8] = b"near-mpc-dilithium-dkg-contribution-v2";
 
 /// Derive a DKG contribution from a master share and tweak.
 ///
@@ -52,43 +56,60 @@ const DKG_CONTRIBUTION_DOMAIN: &[u8] = b"near-mpc-dilithium-dkg-contribution-v1"
 ///
 /// # Security
 ///
-/// The contribution uses SHAKE256 with domain separation. The input includes
-/// the master share's secret key material, ensuring that even though the tweak
-/// is public, only parties holding valid master shares can compute their DKG
-/// contributions.
-///
-/// # Example
-/// ```ignore
-/// use qp_rusty_crystals_threshold::derivation::derive_dkg_contribution;
-///
-/// // Tweak is computed by the caller (e.g., from NEAR MPC contract's derive_dilithium_tweak)
-/// let tweak: [u8; 32] = compute_tweak("alice.near", "ethereum");
-/// let contribution = derive_dkg_contribution(&my_master_share, &tweak);
-///
-/// // Use contribution as seed for DKG randomness
-/// ```
+/// The contribution is computed as
+/// `SHAKE256(domain || party_id || tweak || H(secret_shares))`,
+/// where `secret_shares` is the canonical serialization of every `(subset_mask, s1, s2)`
+/// triple in the share. The shares are the actual cryptographic secret of the threshold
+/// scheme, so an attacker who does not hold a valid `PrivateKeyShare` cannot compute
+/// `derive_dkg_contribution` for any party — even though `party_id`, `tweak`, and
+/// `rho` are public.
 pub fn derive_dkg_contribution(master_share: &PrivateKeyShare, tweak: &[u8; 32]) -> [u8; 32] {
-	// Get secret material from the master share
-	// We use the key field which contains the private key seed
-	let secret_material = master_share.key();
-
-	// Include party_id to ensure different parties get different contributions
-	// even if they somehow had the same key material
 	let party_id_bytes = master_share.party_id().to_le_bytes();
+	let shares_digest = hash_secret_shares(master_share);
 
-	// Use SHAKE256 for key derivation with domain separation
-	// Input: domain || secret || party_id || tweak
 	let mut state = fips202::KeccakState::default();
 	fips202::shake256_absorb(&mut state, DKG_CONTRIBUTION_DOMAIN, DKG_CONTRIBUTION_DOMAIN.len());
-	fips202::shake256_absorb(&mut state, secret_material, secret_material.len());
 	fips202::shake256_absorb(&mut state, &party_id_bytes, party_id_bytes.len());
 	fips202::shake256_absorb(&mut state, tweak, tweak.len());
+	fips202::shake256_absorb(&mut state, &shares_digest, shares_digest.len());
 	fips202::shake256_finalize(&mut state);
 
 	let mut contribution = [0u8; 32];
 	fips202::shake256_squeeze(&mut contribution, 32, &mut state);
 
 	contribution
+}
+
+/// Hash all secret share polynomials into a 64-byte digest for use as keying material.
+///
+/// `BTreeMap` iteration is deterministic by key, so the digest is stable across calls
+/// and across machines holding the same share data.
+pub(crate) fn hash_secret_shares(master_share: &PrivateKeyShare) -> [u8; 64] {
+	let mut state = fips202::KeccakState::default();
+	fips202::shake256_absorb(&mut state, b"threshold-share-digest-v1", 25);
+	fips202::shake256_absorb(&mut state, &master_share.party_id().to_le_bytes(), 4);
+
+	let mut buf: Vec<u8> = Vec::new();
+	for (subset_mask, share_data) in master_share.shares() {
+		buf.clear();
+		buf.extend_from_slice(&subset_mask.to_le_bytes());
+		for poly in &share_data.s1 {
+			for coeff in poly {
+				buf.extend_from_slice(&coeff.to_le_bytes());
+			}
+		}
+		for poly in &share_data.s2 {
+			for coeff in poly {
+				buf.extend_from_slice(&coeff.to_le_bytes());
+			}
+		}
+		fips202::shake256_absorb(&mut state, &buf, buf.len());
+	}
+
+	fips202::shake256_finalize(&mut state);
+	let mut digest = [0u8; 64];
+	fips202::shake256_squeeze(&mut digest, 64, &mut state);
+	digest
 }
 
 /// Identifier for a derived key, used for storage lookup.
@@ -114,19 +135,32 @@ impl DerivedKeyId {
 #[cfg(test)]
 mod tests {
 	use super::*;
-	use crate::participants::ParticipantList;
+	use crate::{keys::SecretShareData, participants::ParticipantList};
 
-	/// Helper to create a test PrivateKeyShare
-	fn create_test_share(party_id: u32, key: [u8; 32]) -> PrivateKeyShare {
+	/// Helper to create a test PrivateKeyShare with synthetic share data.
+	///
+	/// The `key_byte` value is fanned out into both the legacy `key` field and the
+	/// (now security-relevant) `s1`/`s2` polynomial coefficients, so callers that
+	/// previously distinguished shares by their `key` parameter still get distinct
+	/// derivation outputs under the new contribution function.
+	fn create_test_share(party_id: u32, key_byte: u8) -> PrivateKeyShare {
 		let dkg_participants = ParticipantList::new(&[0, 1, 2]).unwrap();
+		let mut shares = std::collections::BTreeMap::new();
+		shares.insert(
+			0b011,
+			SecretShareData {
+				s1: vec![[key_byte as i32; 256]; 7],
+				s2: vec![[key_byte as i32; 256]; 8],
+			},
+		);
 		PrivateKeyShare::new(
 			party_id,
-			3, // total_parties
-			2, // threshold
-			key,
-			[0u8; 32],                         // rho
-			[0u8; 64],                         // tr
-			std::collections::BTreeMap::new(), // shares
+			3,
+			2,
+			[key_byte; 32],
+			[0u8; 32],
+			[0u8; 64],
+			shares,
 			dkg_participants,
 		)
 	}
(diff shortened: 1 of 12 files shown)


n13:
# PR #75 Review: Threshold Cryptography Soundness & Code Issues

I reviewed [PR #75 "Fix issues in threshold 2"](https://github.com/Quantus-Network/qp-rusty-crystals/pull/75) (head `threshold-3` → base `illuzen/threshold-2`). All 198 tests pass locally. Here is what I found.

## Cryptographic Soundness — Verdict: SOUND with caveats

The three claimed fixes (C1, C2, C3) are real and correctly implemented:

### C3 — derived keys (`threshold/src/derivation.rs`)
- `derive_dkg_contribution` now hashes the actual secret share polynomials (`s1`, `s2`) instead of the publicly-derivable `key` field.
- The new test `test_contribution_independent_of_legacy_key_field` (`derivation.rs:228`) correctly proves derivation is independent of `key`. 
- Domain separators properly bumped to `v2` to prevent v1/v2 confusion.
- `hash_secret_shares` (`derivation.rs:87`) uses `BTreeMap` deterministic iteration so the digest is canonical across machines.

### C1 — leak removed (`threshold/src/resharing/types.rs`)
- All five leak vectors (`blinded_s1_contribution`, `blinded_s2_contribution`, `blinding_s1`, `blinding_s2`, `blinding_commitment`) are gone from `ResharingRound1Broadcast`.
- Round 1 now carries only hash commitments. The test `test_round1_broadcast_does_not_leak_subshares` (`protocol.rs:1367`) acts as a tripwire if anyone re-adds plaintext fields.

### C2 — no party reconstructs `s` (`threshold/src/resharing/protocol.rs`)
- The math is correct: for each old subset `I`, dealer `D_I` splits `s_I^old` into sub-shares `r_{I→J}` such that `Σ_J r_{I→J} = s_I^old`. New shares are `s_J^new = Σ_I r_{I→J}`. By swapping summation order: `Σ_J s_J^new = Σ_I s_I^old = s_total` — so the public key `t = A·s1 + s2` is preserved without anyone ever computing `s`.
- Cheating-dealer detection works two ways:
  1. **Recipient-side** (`verify_and_aggregate_new_shares`, `protocol.rs:763`): hashes received `r` and compares to dealer's broadcast commitment.
  2. **Old-subset cross-check** (`collect_accusations`, `protocol.rs:709`): other members of `I` independently recompute `r_{I→J}` from their copy of `s_I^old`.
- The PR's reported bug-fix on `old_subset_order` is real: `compute_old_subset_order` (`protocol.rs:955`) now uses canonical Gosper enumeration over `(n_old, k_old)`, so dealers and verifiers compute identical `residual_idx = i_idx % n_new`.

### Confidentiality / hiding
Hash commitments `H(domain || I || J || r)` are hiding because:
- Non-residual `r` values are sampled from `{-η,…,η}^{(L+K)·N}` ≈ `5^3840` bits — comfortably high min-entropy.
- The residual `r` is `s_I − Σ r_non-residual mod Q`, which inherits its entropy from `s_I` (which is also high-entropy from the attacker's view).

Domain separation across all keyed hashes is clean — every domain string is distinct (`resharing-subset-prf-v2`, `resharing-commit-v2`, `resharing-new-share-commit-v2`, `near-mpc-dilithium-dkg-contribution-v2`, `dkg-party-key-v2`, `reshare-party-key-v2`, `threshold-share-digest-v1`).

### Empirical signing verification
The end-to-end tests don't just check that resharing completes — they sign a real message with the new shares and verify against the **original** public key:

```926:931:threshold/tests/resharing_tests.rs
	let signing_shares: Vec<_> =
		vec![new_shares.get(&0).unwrap().clone(), new_shares.get(&1).unwrap().clone()];
	let is_valid =
		run_signing_and_verify(&signing_shares, &public_key, config, b"test message", b"");

	assert!(is_valid, "Signature with new shares should verify");
```

This passes for `same_committee`, `add_party`, `remove_party`, and `replace_party` scenarios.

## Issues Found

### M1 — Self-message routing relies on network loopback (Medium, robustness)

`build_pending_round2_messages` (`protocol.rs:670`) does **not** skip self when the dealer is also a recipient (i.e., a `Both`-role party in some new subset `J`):

```670:688:threshold/src/resharing/protocol.rs
	fn build_pending_round2_messages(&mut self) {
		let mut by_recipient: BTreeMap<ParticipantId, BTreeMap<SubsetPair, NewShareData>> =
			BTreeMap::new();
		for (pair, share) in &self.my_subshares {
			let j_mask = pair.1;
			for (idx, party) in self.config.new_participants.iter().enumerate() {
				if (j_mask & (1 << idx)) != 0 {
					by_recipient.entry(party).or_default().insert(*pair, share.clone());
				}
			}
		}
```

Then the protocol returns `Action::SendPrivate(self.config.my_party_id, data)`. The receive path (`handle_round2_message`, line 461) only fills `round2_messages` from inbound network deliveries — there's no local short-circuit. So `verify_and_aggregate_new_shares` will fail with *"missing Round 2 message from dealer N"* unless the network layer loops self-private messages back.

The DKG protocol explicitly skips self in `process_round1` (`keygen/dkg/protocol.rs:321`): `if party != state.config.my_party_id`. Resharing should be consistent.

The integration test harness happens to loopback (`resharing_tests.rs:116` doesn't skip `to == party_id`), which is why all 26 resharing tests pass. NEAR MPC's `run_protocol` infrastructure may or may not behave the same way.

**Recommendation:** Either (a) skip self in `build_pending_round2_messages` and locally insert into `round2_messages`, mirroring the DKG pattern; or (b) document the self-loopback requirement in the integration contract.

### M2 — `t = n` configurations have undetected dealer cheating (Pre-existing)

For `t = n`, each subset has size `n − t + 1 = 1`. Each subset `I = {p}` has only one member, so `collect_accusations` (`protocol.rs:709`) can't cross-verify the dealer's commitment — the dealer is the sole member.

A malicious dealer can pick arbitrary `r_{I→J}` (consistent with their own commitments) such that `Σ_J r_{I→J} ≠ s_I^old`. Recipients verify hash matches but cannot detect the wrong sum. The new total secret would not equal the old total, breaking the public key invariant silently.

`ResharingConfig::new` permits `t = n` (only `t < 2` is rejected). A unit test for this case would be valuable — at minimum, document this limitation in `mod.rs`.

### M3 — Test coverage gap: no end-to-end test for fully-disjoint committees

`test_resharing_complete_committee_change` (`resharing_tests.rs:574`) only checks config creation, not the full protocol. There is no E2E signing test for `{0,1,2} → {3,4,5}` (no overlap, all `OldOnly` + `NewOnly`). This is exactly the case where the new committee has zero prior knowledge of the secret, so it would catch any subtle issue in the rho/tr propagation path for `NewOnly` parties (`protocol.rs:899-903`).

### L1 — Liveness: any single offline party halts the protocol (Documented)

- `have_enough_round1` (`protocol.rs:395`) requires broadcasts from **all** old participants, even non-dealers (whose `commitments` map is empty).
- `have_all_round3` (`protocol.rs:572`) requires broadcasts from **all** parties in the union.

Acknowledged in the README under "Limitations" so this is by design — but it means a single offline party DoSes the resharing.

### L2 — DoS via false accusations (Documented)

In `handle_combining` (`protocol.rs:599-611`), any broadcasted `DealerAccusation` triggers an unconditional abort with no adjudication. A single corrupted party can falsely accuse honest dealers and abort. Same for the `success: false` flag.

### L3 — Dead code / cleanup

- `seed: [u8; 32]` (`protocol.rs:191`) is `#[allow(dead_code)]` "reserved for future per-session nonce mixing". If you never plan to use it, drop it; if you do, mix it into `build_subset_seed` so reshares of the same key in different sessions are distinguishable.
- `key` field in `PrivateKeyShare` (`keys.rs:113`) is now flagged dead-for-security (`keys.rs:208`) but kept for serialization. A migration path / version bump should remove it eventually so no one is tempted to use it.
- `verify_new_share_consistency` (`protocol.rs:854-857`) skips the cross-check when `commits.len() < 2`. In the only case this fires (degenerate `t = n` with 1-member subsets), it leaves dealer cheating undetected — see M2.
- `compute_my_subshares` clones `self.old_subset_order` (`protocol.rs:640`) as a borrow-checker workaround. Refactoring with `mem::take` or splitting the borrow would avoid the clone.

### Observation: new shares are NOT η-bounded (intentional, but worth noting)

After resharing, `s_J^new = Σ_I r_{I→J}`, and most `r_{I→J}` are η-bounded but the residual ones are not — they live in `[0, Q)`. This means the per-share η-invariant is broken.

I verified this is **safe**: the signing protocol only uses the *recovered sum* (`recover_share` in `secret_sharing.rs`), and `Σ all shares = s_total mod Q` is preserved exactly. Per-share magnitude is irrelevant once shares are summed mod Q. The signing arithmetic in `generate_round3_response` (`signing.rs:480-538`) operates on the recovered NTT-form `s1`/`s2`, which is identical pre- and post-reshare.

Just be aware: any future code that asserts per-share η-bounds (e.g., a defensive test analogous to `test_mithril_dkg_eta_bounded` from `keygen/dkg/protocol.rs:1115`) will fail on post-reshare shares — that would be a false positive, not a real bug.

## Summary

| Category | Status |
|---|---|
| C1 (Round 1 leak) | Fixed correctly |
| C2 (single-point compromise) | Fixed correctly with sound math |
| C3 (publicly recoverable derived keys) | Fixed correctly |
| Cryptographic soundness | Sound under honest-majority + each old subset has ≥ 1 honest member |
| Self-message routing | Medium robustness issue (M1) |
| `t = n` undetected cheating | Pre-existing soundness gap (M2) |
| Test coverage | Missing E2E for disjoint committees (M3) |
| Code quality | Minor cleanup items (L1–L3) |

The PR is a meaningful security improvement and the cryptographic design is sound. I'd recommend addressing **M1** before merging (small fix, mirrors the DKG pattern) and adding a test for **M3** (one new test function in `resharing_tests.rs`). **M2** is worth at least a documentation note since it's pre-existing.

n13:
All three issues fixed and the test suite is fully green:

| Suite | Before | After |
|---|---|---|
| `--lib` | 128 | **130** (+2 partial-PK unit tests) |
| `resharing_tests` | 26 | **28** (+M3 disjoint-committee + M2 malicious-dealer) |
| `derivation_tests` | 9 | 9 |
| `signer_state_tests` | 16 | 16 |
| `integration_tests` (release) | 19 | 19 |

### What changed

**M1 — Self-message routing (robustness)**
- `threshold/src/resharing/protocol.rs::build_pending_round2_messages` now drops self-deals into `round2_messages` directly and only enqueues real outbound messages, mirroring the DKG pattern.
- `threshold/tests/resharing_tests.rs` test harness now asserts `to != party_id` for every `Action::SendPrivate`, so any future regression that re-introduces self-loopback dependence will fail loudly across all 5 E2E tests.

**M2 — Cheating-dealer detection at `t = n` (cryptographic soundness)**
- New shared module `threshold/src/protocol/partial_pk.rs` with `compute_partial_pk_t` and `pack_combined_pk`, refactored from the DKG (DRY — DKG's old `compute_partial_pk` / `combine_partial_pks` now delegate to it).
- `ResharingRound3Broadcast` gained `partial_pks: BTreeMap<SubsetMask, Vec<[i32; N]>>`, with a serde helper `serde_partial_pks` for `Vec<[i32; 256]>` BTreeMaps.
- `handle_round3_generate` computes `t_J^new = A·s1_J^new + s2_J^new (mod Q)` for every J the party holds.
- `handle_combining` calls a new `verify_public_key_preservation` after the existing consistency check; it cross-checks per-J consensus across new-committee members and packs `Σ_J t_J^new` into the canonical PK encoding for byte-equality against the original public key. Mismatch ⇒ `ShareVerificationFailed`.
- Hardened by `test_resharing_detects_consistent_dealer_tamper_at_t_equals_n`, which forges a *consistent* `(commitment, r)` pair via `forge_consistent_commitment` so the recipient's commitment-vs-r check passes — and then asserts the resharing fails with the public-key invariant error. Two new `partial_pk` unit tests cover additivity and tamper-detection of the underlying math.

**M3 — Coverage gap**
- `test_resharing_end_to_end_disjoint_committees` runs `(2,3) {0,1,2} → (2,3) {3,4,5}` with no overlap, then signs against the original public key with two different threshold subsets of the brand-new committee.

Quick housekeeping along the way: removed three unused DKG-local constants (`Q`, `PUBLIC_KEY_SIZE`, `TR_SIZE`) and unused imports (`packing`, `polyvec`, DKG `K`/`N`) that became dead after the partial-PK refactor; `build_private_key_share` now reuses the new `derive_rho` helper. README updated to describe the partial-PK invariant check, the local self-deal routing, and the new `Round3Broadcast` field.

PR #77, opened 2026-05-18
Set max msg size to prevent OOM panics

illuzen:
Simple validation, someone could crash the chain by supplying a too-big message. If a message this big needs to be signed they can hash then sign.

---

> [!NOTE]
> **Medium Risk**
> Adds a hard 64MiB input limit to cryptographic `sign`/`verify`, which can cause new rejections for previously-accepted large messages but reduces DoS/OOM risk.
> 
> **Overview**
> Adds a **64 MiB maximum message size** (`MAX_MESSAGE_SIZE`) for ML-DSA-87 signing and verification to prevent memory-exhaustion/OOM scenarios.
> 
> `SecretKey::sign` now returns `SignatureError::MessageTooLong` when exceeded, and `PublicKey::verify` returns `false` early for oversized messages; docs are updated and tests cover both rejection paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d21455a8029f844ec785dc1bdbeefac2a1b8abda. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dilithium/src/errors.rs
+++ b/dilithium/src/errors.rs
@@ -21,12 +21,14 @@ impl Display for KeyParsingError {
 #[derive(Debug)]
 pub enum SignatureError {
 	ContextTooLong,
+	MessageTooLong,
 }
 
 impl Display for SignatureError {
 	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
 		let str = match self {
 			SignatureError::ContextTooLong => "ContextTooLong",
+			SignatureError::MessageTooLong => "MessageTooLong",
 		};
 		write!(f, "{str}")
 	}
--- a/dilithium/src/ml_dsa_87.rs
+++ b/dilithium/src/ml_dsa_87.rs
@@ -81,9 +81,14 @@ impl Keypair {
 	///
 	/// # Arguments
 	///
-	/// * 'msg' - message to sign
+	/// * 'msg' - message to sign (max 64 MiB)
+	/// * 'ctx' - optional context string (max 255 bytes)
+	/// * 'hedge' - optional random bytes for hedged signing
 	///
-	/// Returns Result<Signature, SignatureError>
+	/// # Errors
+	///
+	/// Returns `SignatureError::MessageTooLong` if the message exceeds 64 MiB.
+	/// Returns `SignatureError::ContextTooLong` if the context exceeds 255 bytes.
 	pub fn sign(
 		&self,
 		msg: &[u8],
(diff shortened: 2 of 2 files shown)


n13:
Here's my review of PR #77:

---

## PR #77: Set max msg size to prevent OOM panics

**Summary**: Adds a 64 MiB `MAX_MESSAGE_SIZE` limit to `SecretKey::sign` and `PublicKey::verify`, guarding against a DoS vector where an attacker supplies a massive message to trigger OOM via the `vec![0; msg_len + ...]` allocations in both methods.

### What looks good

- **Correct guard placement.** The check in `sign` is placed *before* the `match ctx` block, guarding both the `Some` and `None` branches where the oversized `vec!` allocation would occur. Same for `verify` — it sits right after the cheap `sig.len()` check and before any allocation.

- **Reasonable limit.** 64 MiB is generous enough for any realistic Dilithium signing use case (transactions, hashes, etc.) while capping the damage an attacker can do. The `>` comparison correctly allows exactly 64 MiB.

- **Doc cleanup.** The old `sign` doc said `Returns Option<Signature>` when it actually returns `Result` — this PR fixes that and properly documents the error variants.

- **Delegation is covered.** `Keypair::sign` and `Keypair::verify` both delegate to the protected methods, so no bypass path exists.

### Concerns

1. **No tests for the new paths.** There are no tests asserting that a message over the limit produces `SignatureError::MessageTooLong` on sign, or `false` on verify. This is a security boundary — it should have test coverage. Something like:

```rust
#[test]
fn sign_rejects_oversized_message() {
    let keys = Keypair::generate(get_random_bytes());
    let big_msg = vec![0u8; MAX_MESSAGE_SIZE + 1];
    let result = keys.sign(&big_msg, None, None);
    assert!(matches!(result, Err(SignatureError::MessageTooLong)));
}

#[test]
fn verify_rejects_oversized_message() {
    let keys = Keypair::generate(get_random_bytes());
    let big_msg = vec![0u8; MAX_MESSAGE_SIZE + 1];
    assert!(!keys.verify(&big_msg, &[0u8; SIGNBYTES], None));
}
```

2. **Asymmetric error reporting.** `sign` returns a typed `Err(SignatureError::MessageTooLong)` while `verify` silently returns `false`. This is consistent with the existing `verify` API (it already returns `bool`, and the context-too-long and sig-length checks also return `false`), so I don't think it needs to change *now*. But it's worth noting that callers can't distinguish "bad signature" from "message too large" on the verify side. If diagnostics matter downstream, this could be revisited later with a `Result`-returning verify variant.

3. **Minor: `Keypair` doc comments are now stale.** The `Keypair::sign` doc still says `Returns Result<Signature, SignatureError>` without mentioning `MessageTooLong`, and `Keypair::verify` doesn't mention the size limit. Since these delegate directly to the updated methods, it'd be nice to keep them in sync — or just remove the redundant docs on the delegating methods.

### Verdict

Clean, focused, and addresses a real DoS vector. The only blocking item I'd flag is the missing tests — everything else is minor.

n13 (approved):
Approved

Tests could be added

The rest of the agent is not a bug or not important

PR #78, opened 2026-05-21
Rustier crystals

illuzen:
## Summary

Refactor `dilithium` crate from C-like to idiomatic Rust with type-safe APIs (Tier 2) to address audit findings about unchecked slice indexing and panic vectors.

### Type Safety
- All internal functions now use fixed-size arrays instead of slices with separate length parameters
- Compile-time enforcement of buffer sizes eliminates runtime bounds check failures
- Public API boundary (`ml_dsa_87`) handles slice-to-array conversion gracefully

### Module Visibility  
- Internal modules (`fips202`, `ntt`, `poly`, `polyvec`, `packing`, `reduce`, `rounding`, `sign`) now `pub(crate)`
- Only `ml_dsa_87`, `params`, `SensitiveBytes32`/`SensitiveBytes64`, and error types are public

### Audit Issues Addressed
| Issue | Resolution |
|-------|------------|
| Unchecked slice indexing | Fixed-size arrays enforce bounds at compile time |
| `rej_eta` modulo-by-zero potential | Explicit early return for empty output |
| Redundant length parameters | Removed - functions use slice `.len()` directly |
| Internal functions exposed | Modules now `pub(crate)` |

### Code Cleanup
- Removed unused functions: `poly::add`, `poly::sub`, `poly::use_hint_ip`
- Removed `ct_bench.rs` example and `internal` feature
- Zero compiler warnings, zero clippy lints

## Testing
- 81 unit tests pass
- 10 integration tests pass (including NIST KAT vectors)
- All tests pass in debug and release mode

## Breaking Changes
None - public API unchanged. Internal module visibility reduced but these were not part of the public API contract.

---

> [!NOTE]
> **Medium Risk**
> Touches core signature/hash/packing code paths and changes many function signatures to fixed-size arrays; while largely type-safety refactors, any mismatch in buffer handling could affect cryptographic correctness or interoperability.
> 
> **Overview**
> Refactors the `dilithium` crate’s internals to use **fixed-size array buffers** instead of `&[u8]` + explicit length parameters, pushing length validation to the public `ml_dsa_87` boundary (e.g., signature verification now converts `sig` via `try_into()` before calling internal verify logic).
> 
> Updates Keccak/SHAKE (`fips202`) and related crypto plumbing (`sign`, `poly`, `polyvec`, `packing`, `ntt`) to match the new type-safe APIs, including safer block processing (`chunks_exact`) and adding NIST SHAKE256 vector tests plus an explicit `rej_eta` empty-output guard. Removes unused polynomial helpers and drops the `ct_bench` example, and adds a `clippy.sh` helper to mirror CI clippy settings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ea498cca107f7680bc89a80553fb29d5bf0bc7e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13:
I've reviewed the PR end-to-end: read the full diff (10 files, +207/−1452), inspected each module's before/after, traced the affected call paths, checked workspace-wide callers for the visibility tightening, and built + ran the full test suite (`cargo test --release` → 81 unit + 10 integration + KAT all pass, `cargo clippy --all-targets -- -D warnings` clean) on the PR branch.

## Overall verdict: LGTM with a few notes

This is a well-scoped, mostly mechanical Tier-2 refactor that meaningfully shrinks the panic surface of the dilithium crate without changing the public API. Net deletion of ~1.2k lines is a real win.

## What's good

**Type-safe internals.** Switching internal Keccak/SHAKE, NTT, packing, poly, polyvec, sign, and verify from `&[u8]` + redundant `len` parameter to fixed-size arrays (`&[u8; SEEDBYTES]`, `&[u8; SECRETKEYBYTES]`, `&[u8; SIGNBYTES]`, etc.) genuinely removes runtime bounds-check / panic surface and pushes errors to compile time. `redundant length parameter` removal is also nice — the previous `inlen`/`outlen`/`alen` arguments were trust-the-caller foot-guns.

**Public API stays binary-compatible.** `ml_dsa_87::{Keypair, PublicKey, SecretKey}` and their `to_bytes` / `from_bytes` / `sign` / `verify` shapes are unchanged. The new `PublicKey::verify` length check via `try_into` is semantically identical to the old `if sig.len() != SIGNBYTES { return false; }` — it just produces the typed reference the internal `verify` now requires:

```237:243:dilithium/src/ml_dsa_87.rs
	pub fn verify(&self, msg: &[u8], sig: &[u8], ctx: Option<&[u8]>) -> bool {
		// Validate signature length first
		let sig: &[u8; SIGNBYTES] = match sig.try_into() {
			Ok(s) => s,
			Err(_) => return false,
		};
```

**Module visibility tightening is safe.** I grepped the whole workspace (`hdwallet`, `tests`, benches, examples, FFI dir) for `qp_rusty_crystals_dilithium::(fips202|ntt|packing|poly|polyvec|reduce|rounding|sign)` — zero external references. Demoting these to `pub(crate)` doesn't break any in-tree consumer, and they were not part of the documented public API anyway.

**`rej_eta` empty-output fix is correct.** This is a real (if unreachable-in-practice) bug. The function does `a[ctr % alen]` — with `alen == 0` that is `% 0`, which panics. The early return prevents it:

```354:362:dilithium/src/poly.rs
pub fn rej_eta(a: &mut [i32], buf: &[u8]) -> usize {
	let alen = a.len();
	let buflen = buf.len();

	// Early return for empty output - avoids any potential issues with modulo operations
	if alen == 0 {
		return 0;
	}
```

I traced `uniform_eta`'s outer/inner loop arithmetic and confirmed `alen=0` can't be hit in normal signing (`total_coefficients_collected` is bounded by `N=256` before the outer loop exits, while the storage is `[i32; 1000]`), so this is defensive rather than fix-of-live-bug, but absolutely worth having.

**Bonus latent bug squashed in `keccak_squeeze`.** Master had this in the multi-iteration outer-loop case:

```12:25:dilithium/src/fips202.rs
	while outlen != 0 {
		if pos == r {
			keccakf1600_statepermute(s);
			pos = 0;
		}
		let mut i = pos;
		let mut idx = 0;   // <- reset on every outer iteration
		while i < r && i < pos + outlen {
			out[idx] = ...;
```

If `outlen > r`, the second outer iteration would reset `idx` to 0 and overwrite the first block's output. In dilithium every caller asks for `outlen < r` (CRHBYTES=64, C_DASH_BYTES=64, both < SHAKE256_RATE=136), so it never fired — but the PR's new version hoists `out_idx` outside the outer loop, quietly fixing it. Worth calling out in the PR description so future spelunkers don't think it's intentional.

**Removed-as-unused are genuinely unused.** `poly::add`, `poly::sub` had no callers. `poly::use_hint_ip` was a byte-for-byte duplicate of `poly::use_hint` (the `_ip` suffix was a lie — `use_hint` is also in-place). The stale `Returns polynomial with corrected high bits` doc on `use_hint` (which returns `()`) is also cleaned up. Good.

**Tests.** New `test_rej_eta_empty_output` covers the new branch; the old `test_invalid_signature_length` was moved up to where the validation now lives (with an explanatory comment). 81/0 unit + 10/0 integration green, including the NIST KAT vector test.

## Things worth pushing back on / questions

1. **`ct_bench.rs` removal is the biggest concrete loss.** Deleting ~1.2k lines of dudect-bencher constant-time tests is most of the diff's deletion count, but it's also deleting a real audit instrument. The summary frames this as cleanup, but timing-side-channel coverage isn't free to rebuild. A couple of options worth considering before merging:
   - Keep `ct_bench` in tree (it's only an example, doesn't cost runtime users anything).
   - Or move it under a `ct_bench` feature flag.
   - Or at minimum, confirm `dilithium/CONSTANT_TIME_TESTING.md` is updated / not now lying about how to run these.

   If the answer is "we have a different CT harness in another repo" — fine, but the PR description should say so.

2. **`unwrap()` in `keccak_absorb_once` / `keccak_squeezeblocks` is implicit-panic territory.** The new code does:

   ```374:381:dilithium/src/fips202.rs
   	while inlen >= r {
   		for i in 0..r / 8 {
   			let bytes: &[u8; 8] = input[idx + 8 * i..idx + 8 * i + 8].try_into().unwrap();
   			s[i] ^= load64(bytes);
   ```

   This is reachable only with internally-mis-sized buffers (callers all pass either fixed arrays or `nblocks * r`-sized slices, so it's currently safe). Given the whole point of the PR is "no unchecked indexing", these two functions could finish the job by taking `&[u8; SHAKE128_RATE]` / `&[u8; SHAKE256_RATE]` (or chunking with `chunks_exact`) instead of `&[u8]` + `r: usize`. Minor — not a blocker — but it would close out the same audit class fully.

3. **`shake256(output, input)` no longer takes explicit `outlen`/`inlen`** — using `output.len()` and `input.len()` is fine, but worth being deliberate: any caller that previously passed a smaller `outlen` than `output.len()` (none in this repo, I checked) would silently change behavior. Just confirm there's no downstream consumer doing that on a pre-allocated-larger buffer pattern.

4. **Doc comment "Always process exactly buflen bytes" on `rej_eta`** is now slightly misleading given the new `alen == 0` early return. The CT property still holds for the only realistic input set (`alen > 0`), but a one-line note would prevent confusion.

5. **`packing::pack_sig` still accepts `c: Option<&[u8; C_DASH_BYTES]>`.** Looking at the two call sites, one always passes `Some(...)` (final pack) and the other always passes `None` (dummy/constant-time pack). The `Option` is effectively a "skip the first C_DASH_BYTES" toggle, which is a slightly awkward API. Not introduced by this PR — pre-existing — but if you're touching the signature anyway, splitting into `pack_sig_with_challenge` / `pack_sig_no_challenge` would read better. Optional.

6. **Stylistic / per-repo-rules nits** (also not blockers):
   - The new `// Validate signature length first` comment in `PublicKey::verify` is the kind of narrate-what-the-code-does comment that's typically discouraged.
   - `// Early return for empty output - avoids any potential issues with modulo operations` is fine because it explains *why*, not *what*.

## Recommendation

Approve once the `ct_bench` deletion is justified (kept elsewhere, replaced, or intentionally retired). Everything else above is non-blocking polish. The core refactor is sound, the test suite proves equivalence, the audit goals are met, and the latent `keccak_squeeze` bug fix is a nice freebie.

n13:
# PR #78 Review — `Rustier crystals`

Reviewed the full diff (13 files, +1836/−1468 vs merge base `6c06374`), ran the full workspace test suite and the CI clippy command on the PR branch, and dry-merged `origin/master` (which now has #79) into it.

## Verification

| Check | Result |
|---|---|
| `cargo build -p qp-rusty-crystals-dilithium` | clean |
| `bash clippy.sh` (= `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`) | clean |
| `cargo test -p qp-rusty-crystals-dilithium --release` | 83 passed (n13 saw 81 — the 2 new NIST SHAKE256 tests account for the delta) |
| `cargo test --release` (full workspace) | 83 + 34 hdwallet + 10 integration (incl. NIST KAT) all pass |
| `git merge origin/master` into PR | auto-resolves cleanly, tests still 83/0 after merge |
| Grep for `unwrap()`/`expect(` in non-test code | only the pre-existing `expect("Should never fail")` in `Keypair::generate` |

## What's good

- **Type safety win is real.** The Keccak/SHAKE/NTT/packing/poly/polyvec/sign internals now consume `&[u8; SEEDBYTES]`, `&[u8; CRHBYTES]`, `&[u8; SECRETKEYBYTES]`, `&mut [u8; SIGNBYTES]`, `&mut [i32; N]`, etc. instead of `&[u8]` + redundant `inlen`/`outlen`/`alen`. The redundant-length params were classic trust-the-caller foot-guns — good riddance.
- **Public API stays binary-compatible.** `ml_dsa_87::{Keypair,PublicKey,SecretKey}` keep their `to_bytes/from_bytes/sign/verify` shapes. The new `try_into()` in `PublicKey::verify` is semantically identical to the old `if sig.len() != SIGNBYTES { return false; }`.
- **Module visibility tightening is safe.** Grepped the workspace for `qp_rusty_crystals_dilithium::(fips202|ntt|packing|poly|polyvec|reduce|rounding|sign)` — no external in-tree callers. The demotion to `pub(crate)` doesn't break anyone, and `errors::{KeyParsingError, SignatureError}` are now re-exported from the crate root which is a nice fixup.
- **`rej_eta` empty-output early return** prevents the latent modulo-by-zero (`a[ctr % alen]` with `alen == 0`). Defensive but worth having, and the comment explains *why* rather than *what*:

```361:373:dilithium/src/poly.rs
pub fn rej_eta(a: &mut [i32], buf: &[u8]) -> usize {
	let alen = a.len();
	let buflen = buf.len();

	// Early return for empty output - avoids any potential issues with modulo operations
	if alen == 0 {
		return 0;
	}
```

- **Bonus latent bug fix in `keccak_squeeze`.** Master had `let mut idx = 0;` inside the outer `while outlen != 0` loop, which would reset the output index on every iteration and overwrite the first block. Only unfired in practice because every dilithium caller squeezes `< SHAKE256_RATE` bytes, but the PR hoists `out_idx` outside and quietly fixes it. Worth a one-liner in the PR description so a future spelunker doesn't think it was intentional.
- **`load64` is type-safe and inlined.** Now `fn load64(x: &[u8; 8]) -> u64 { u64::from_le_bytes(*x) }` — same codegen, no bounds checks, no `pub`.
- **`keccak_absorb_once` now uses `chunks_exact`.** The previous `unwrap()`s n13 flagged in their review are gone — addressed in commit `86b821a` ("ditch unwraps, improve comments").
- **NIST SHAKE256 KAT vectors are a great addition.** Both Short and Long message files are run end-to-end through the new absorb/finalize/squeeze path — this is strong evidence that the refactor preserves cryptographic correctness, not just compiles.
- **`poly::add`/`poly::sub` removal is correct** — confirmed they had no callers in the workspace.
- **`poly::use_hint_ip` was a literal byte-for-byte duplicate of `poly::use_hint`** (the `_ip` suffix was a lie — both are in place). Removal cleans up a real DRY violation and removes the stale `Returns polynomial with corrected high bits` docstring that lied about a `()`-returning fn.
- **`clippy.sh` mirroring CI** — nice DX.
- **No merge conflicts with #79** even though both touch `lib.rs`, `ml_dsa_87.rs`, `sign.rs`.

## Concerns

### Should be addressed before merge

1. **`ct_bench.rs` deletion (−1195 lines) needs explicit justification.** This is most of the diff's deletion count and it's deleting a real audit instrument (dudect-bencher constant-time tests). The PR description frames this as "cleanup" but timing-side-channel coverage isn't free to rebuild. Options to consider:
   - keep `ct_bench` in tree (it's only an example — no cost to runtime users),
   - move it behind a `ct_bench` feature,
   - or, if there's a different CT harness in another repo, say so in the PR body.
   Also worth confirming `dilithium/CONSTANT_TIME_TESTING.md` isn't now stale. This is the only item I'd actually block on.

### Worth fixing, not blocking

2. **`keccak_squeezeblocks` silently caps via `take(nblocks)`.** This violates the repo's "fail early — never have a silent failure" rule:

```376:385:dilithium/src/fips202.rs
fn keccak_squeezeblocks(out: &mut [u8], nblocks: usize, s: &mut [u64; 25], r: usize) {
	// Process exactly nblocks blocks using chunks_exact_mut for safe iteration
	for block in out.chunks_exact_mut(r).take(nblocks) {
		keccakf1600_statepermute(s);
		for (i, chunk) in block.chunks_exact_mut(8).enumerate() {
			// chunks_exact_mut(8) guarantees exactly 8 bytes
			let bytes = s[i].to_le_bytes();
			chunk.copy_from_slice(&bytes);
		}
	}
```

   If a caller ever passes `out` smaller than `nblocks * r`, this silently does fewer squeezes — bytes the caller expects are uninitialized, and signing/verification will produce wrong results with no error. I verified all 7 current call sites are statically sized adequately, so it's not a live bug, but a single `debug_assert!(out.len() >= nblocks * r)` (or even a `let blocks_done = ...; assert_eq!(blocks_done, nblocks);` after the loop) would catch a future regression at the boundary the PR is otherwise trying to make compile-time safe.

3. **`keccak_absorb` still uses unchecked `input[idx]` indexing.** This is the one remaining function that didn't get the chunks-based treatment — the loop bounds prevent it from panicking in practice, but it's the same audit class the PR was meant to close. Since it's the incremental absorb (variable input length) it can't take a fixed array directly, but it could be rewritten the same way `keccak_absorb_once` was. Not a blocker, but it'd finish the job.

4. **`k_pack_w1` still takes `&mut [u8]`** not a fixed array, even though every caller passes a buffer that's exactly `K * params::POLYW1_PACKEDBYTES` (or a known-larger prefix of `&mut [u8; SIGNBYTES]`). Pre-existing style inconsistency given the rest of the PR. Easy follow-up.

### Pure nits

5. **`// Validate signature length first`** in `PublicKey::verify` is exactly the kind of narrate-what-the-code-does comment the repo's rules ask to avoid. The `try_into()` self-documents.
6. **The `SAFETY` comment in `keccak_absorb_once`** is followed by `chunk[0], chunk[1], ..., chunk[7]` direct indexing. It works but defeats bounds-check elimination — `let bytes: &[u8; 8] = chunk.try_into().expect("chunks_exact(8) yields 8 bytes");` is the idiomatic equivalent and unwraps to a no-op cast on a `chunks_exact` iterator.
7. **`pack_sig`'s `c: Option<&[u8; C_DASH_BYTES]>`** is a slightly awkward "skip first C_DASH_BYTES" toggle. Pre-existing, not introduced by this PR, but splitting into `pack_sig_with_challenge` / `pack_sig_no_challenge` would read better if you're touching the signature anyway. Optional.
8. Doc on `rej_eta` says *"Always process exactly buflen bytes"* in the body comment — still technically true for the only realistic input set (`alen > 0`), but a one-liner acknowledging the new early-return wouldn't hurt.

## Verdict

**Approve once the `ct_bench` deletion is justified** in the PR description (or the file is kept/feature-gated). Everything else above is non-blocking polish. Core refactor is sound, audit goals are met, KAT + NIST SHAKE256 vectors prove equivalence, and the `keccak_squeeze` latent fix is a nice freebie. This aligns with the prior review by @n13 ([Audit-driven dilithium refactor review](b3f01a5a-1234-5678-9012-3456789abcde)).

Want me to draft any of the polish items as follow-up commits on the PR branch, or leave them as comments?

n13 (approved):
It's in good shape now!

PR #79, opened 2026-05-21
fixing audit issues

n13:
All three security findings are fixed, tested, and clippy-clean. Quick summary:

**Issue 1 — Unbounded derivation-path DoS** (`[hdwallet/src/lib.rs](hdwallet/src/lib.rs)`, `[hdwallet/src/hderive.rs](hdwallet/src/hderive.rs)`)
- New public constants `MAX_DERIVATION_DEPTH = 16` and `MAX_DERIVATION_PATH_BYTES = 256`.
- New explicit errors `HDLatticeError::PathTooLong(usize)` / `PathTooDeep(usize)` and matching `hderive::Error::PathTooLong` / `PathTooDeep` — fail-early, no silent truncation.
- Shared bounds check `hderive::check_path_bounds` runs first in `DerivationPath::from_str` (defense in depth) and again at `check_derivation_path` before any allocation/HMAC work. DRY: one function, two callers.
- `generate_wormhole_from_seed` now calls `check_derivation_path` before its `nth(2)` check.
- Three new tests cover oversized path, too-deep path, and exact-max-depth boundary, for both `derive_key_from_seed` and `generate_wormhole_from_seed`.

**Issue 2 — Mnemonic not zeroized on parse failure** (`[hdwallet/src/lib.rs](hdwallet/src/lib.rs)`, `[hdwallet/Cargo.toml](hdwallet/Cargo.toml)`)
- `mnemonic_to_seed` now wraps the input in `zeroize::Zeroizing<String>` immediately, so the heap buffer is zeroized on every exit path (success, parse error, panic unwind).
- Enabled `alloc` feature on `zeroize` to bring in the `String: Zeroize` impl.
- New test `test_mnemonic_to_seed_zeroizes_on_parse_failure` exercises the parse-error path.

**Issue 3 — Clone on move-only sensitive types** (`[dilithium/src/lib.rs](dilithium/src/lib.rs)`, `[hdwallet/src/hderive.rs](hdwallet/src/hderive.rs)`, `[dilithium/src/sign.rs](dilithium/src/sign.rs)`)
- Removed `Clone` from `SensitiveBytes32`, `SensitiveBytes64`, and `ExtendedPrivKey`. Doc-comments updated to make the move-only contract a compile-time guarantee.
- The single in-tree call site that broke (`test_fixed_seed_keypair`) now builds two fresh `SensitiveBytes32` instances from the same fixed bytes, which is the test's actual intent.

**Out-of-scope follow-up (flagged in plan, not implemented):** `Keypair`, `SecretKey`, and `WormholePair` still derive `Clone` over secret material and would benefit from the same treatment when ready for a public-API version bump.

Validation: `./run_all_tests.sh` → 34 hdwallet + 10 integration + dilithium unit + doc tests all green; `cargo clippy --workspace --all-targets -- -D warnings` clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,12 @@ This workspace provides post-quantum cryptographic primitives and HD wallet func
 
 ## Security Features
 
-This implementation provides enterprise-grade memory security through `SensitiveBytes*` wrapper types that:
+This implementation provides enterprise-grade memory security:
 
-- **Prevent accidental copying** - Compile-time enforcement ensures sensitive data can only be moved, never copied
-- **Automatic memory zeroization** - Both source arrays and wrapper contents are automatically cleared when dropped
-- **Explicit sensitive data handling** - API requires `(&mut entropy).into()` syntax, making sensitive operations obvious
-- **Move-only semantics** - Functions take `SensitiveBytes32`/`SensitiveBytes64` directly, preventing silent reference copying
+- **Compile-time copy prevention on secrets** - `SensitiveBytes32`, `SensitiveBytes64`, `Keypair`, `SecretKey`, `ExtendedPrivKey`, and `WormholePair` deliberately do **not** implement `Clone`. Secret material can only be moved. Any duplication must be explicit via `to_bytes()`/`from_bytes()` round-trips, making the copy visible at the call site.
+- **Automatic memory zeroization** - Source arrays passed to constructors and all wrapper contents are zeroed on drop (`ZeroizeOnDrop`). Mnemonic strings passed to `mnemonic_to_seed` are wrapped in `Zeroizing<String>` and scrubbed on **every** exit path, including parse failure and panic unwind.
+- **Explicit sensitive data handling** - APIs require `(&mut entropy).into()` syntax, making sensitive operations obvious and zeroing the caller's buffer at the point of transfer.
+- **Bounded derivation paths** - `derive_key_from_seed` and `generate_wormhole_from_seed` reject paths whose byte length exceeds `MAX_DERIVATION_PATH_BYTES` (256) or whose segment count exceeds `MAX_DERIVATION_DEPTH` (16) **before** any allocation or HMAC work, preventing DoS via attacker-controlled deep paths.
 
 ```rust
 // Secure by design - entropy is zeroized after conversion
@@ -19,6 +19,11 @@ let mut entropy = [0u8; 32];
 getrandom::getrandom(&mut entropy).unwrap();
 let keypair = ml_dsa_87::Keypair::generate((&mut entropy).into());
 // entropy is now [0,0,0,...] - no sensitive data left in memory
+
+// If you genuinely need a second copy of a keypair, you must opt in explicitly:
+let kp_bytes = keypair.to_bytes();
+let keypair2 = ml_dsa_87::Keypair::from_bytes(&kp_bytes)?;
+// Prefer passing &Keypair instead of duplicating secrets.
 ```
 
 This eliminates entire classes of security vulnerabilities related to sensitive data handling in cryptographic applications.
--- a/dilithium/examples/stack_usage_demo.rs
+++ b/dilithium/examples/stack_usage_demo.rs
@@ -139,6 +139,7 @@ fn main() {
 	// Pre-generate test data for all variants
 	let entropy = get_random_bytes();
 	let ml87_keypair = ml_dsa_87::Keypair::generate(entropy);
+	let ml87_keypair_bytes = ml87_keypair.to_bytes();
 
 	let test_msg = b"stack usage test message";
 
@@ -169,16 +170,18 @@ fn main() {
 			true
 		});
 
-		let ml87_keypair_clone = ml87_keypair.clone();
 		let ml87_sign = test_sign_with_stack_size(size_kb, "ml-dsa-87", move || {
-			let _sig = ml87_keypair_clone.sign(test_msg, None, None);
+			let kp = ml_dsa_87::Keypair::from_bytes(&ml87_keypair_bytes)
+				.expect("from_bytes must succeed for known-good bytes");
+			let _sig = kp.sign(test_msg, None, None);
 			true
 		});
 
-		let ml87_keypair_clone2 = ml87_keypair.clone();
 		let ml87_sig_clone = ml87_sig;
 		let ml87_verify = test_verify_with_stack_size(size_kb, "ml-dsa-87", move || {
-			ml87_keypair_clone2.verify(test_msg, &ml87_sig_clone, None)
+			let kp = ml_dsa_87::Keypair::from_bytes(&ml87_keypair_bytes)
+				.expect("from_bytes must succeed for known-good bytes");
+			kp.verify(test_msg, &ml87_sig_clone, None)
 		});
 
 		println!(
--- a/dilithium/src/lib.rs
+++ b/dilithium/src/lib.rs
@@ -12,7 +12,9 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 /// Wrapper for sensitive 32-byte data that enforces move semantics and automatic zeroization
 ///
 /// Both `new()` and `from()` take mutable references and zeroize the input data,
-/// ensuring no copies of sensitive data remain in memory.
+/// ensuring no copies of sensitive data remain in memory. `Clone` is intentionally
+/// not derived: the type is move-only by construction, so the secret can exist in
+/// at most one wrapper instance at a time.
 ///
 /// ```rust
 /// use qp_rusty_crystals_dilithium::SensitiveBytes32;
@@ -21,7 +23,7 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 /// // or
 /// let sensitive = SensitiveBytes32::from(&mut entropy); // same behavior
 /// ```
-#[derive(Clone, ZeroizeOnDrop)]
+#[derive(ZeroizeOnDrop)]
 pub struct SensitiveBytes32([u8; 32]);
 
 impl SensitiveBytes32 {
@@ -52,8 +54,10 @@ impl From<&mut [u8; 32]> for SensitiveBytes32 {
 /// Wrapper for sensitive 64-byte data that enforces move semantics and automatic zeroization
 ///
 /// Both `new()` and `from()` take mutable references and zeroize the input data,
-/// ensuring no copies of sensitive data remain in memory.
-#[derive(Clone, ZeroizeOnDrop)]
+/// ensuring no copies of sensitive data remain in memory. `Clone` is intentionally
+/// not derived: the type is move-only by construction, so the secret can exist in
+/// at most one wrapper instance at a time.
+#[derive(ZeroizeOnDrop)]
 pub struct SensitiveBytes64([u8; 64]);
 
 impl SensitiveBytes64 {
--- a/dilithium/src/ml_dsa_87.rs
+++ b/dilithium/src/ml_dsa_87.rs
@@ -21,7 +21,11 @@ pub const MAX_MESSAGE_SIZE: usize = 64 * 1024 * 1024;
 pub type Signature = [u8; SIGNBYTES];
 
 /// A pair of private and public keys.
-#[derive(Clone)]
+///
+/// `Clone` is intentionally not derived because the embedded `SecretKey` is sensitive.
+/// To explicitly copy a keypair (e.g. to move it into a closure), serialize and
+/// reconstruct: `Keypair::from_bytes(&keypair.to_bytes())?`. This forces the
+/// duplication of secret material to be visible at every call site.
 pub struct Keypair {
 	pub secret: SecretKey,
 	pub public: PublicKey,
@@ -121,7 +125,11 @@ impl fmt::Debug for Keypair {
 }
 
 /// Private key.
-#[derive(Clone, ZeroizeOnDrop)]
+///
+/// `Clone` is intentionally not derived because the underlying bytes are sensitive.
+/// To explicitly copy a secret key, use `SecretKey::from_bytes(&sk.to_bytes())?`,
+/// which makes the duplication of secret material visible at every call site.
+#[derive(ZeroizeOnDrop)]
 pub struct SecretKey {
 	pub bytes: [u8; SECRETKEYBYTES],
 }
--- a/dilithium/src/sign.rs
+++ b/dilithium/src/sign.rs
@@ -700,17 +700,16 @@ mod tests {
 
 	#[test]
 	fn test_fixed_seed_keypair() {
-		let seed = get_random_bytes();
+		let seed_bytes = [0x55u8; crate::params::SEEDBYTES];
 
 		let mut pk1 = [0u8; crate::params::PUBLICKEYBYTES];
 		let mut sk1 = [0u8; crate::params::SECRETKEYBYTES];
 		let mut pk2 = [0u8; crate::params::PUBLICKEYBYTES];
 		let mut sk2 = [0u8; crate::params::SECRETKEYBYTES];
 
-		super::keypair(&mut pk1, &mut sk1, seed.clone());
-		super::keypair(&mut pk2, &mut sk2, seed);
+		super::keypair(&mut pk1, &mut sk1, (&mut seed_bytes.clone()).into());
+		super::keypair(&mut pk2, &mut sk2, (&mut seed_bytes.clone()).into());
 
-		// Same seed should produce same keypair
 		assert_eq!(pk1, pk2);
 		assert_eq!(sk1, sk2);
 	}
--- a/hdwallet/Cargo.toml
+++ b/hdwallet/Cargo.toml
@@ -25,7 +25,7 @@ hex-literal.workspace = true
 serde_json.workspace = true
 hex.workspace = true
 serde.workspace = true
-zeroize.workspace = true
+zeroize = { workspace = true, features = ["alloc"] }
 
 # Dependencies for hderive module (ported from nam-tiny-hderive)
 hmac = { version = "0.12", default-features = false }
--- a/hdwallet/src/hderive.rs
+++ b/hdwallet/src/hderive.rs
@@ -3,16 +3,31 @@ use core::{ops::Deref, str::FromStr};
 use hmac::{Hmac, Mac};
 use sha2::Sha512;
 
-use crate::SensitiveBytes32;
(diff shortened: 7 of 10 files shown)


n13:
## PR review: [fixing audit issues #79](https://github.com/Quantus-Network/qp-rusty-crystals/pull/79)

Reviewed the diff against the three audit write-ups (`sec-1.txt`, `sec-2.txt`, `sec-3.txt`). The four new hdwallet tests pass locally. Overall this is a solid fix set; I would approve with a few follow-ups noted below.

---

### Verdict by finding

| Finding | Addressed? | Notes |
|--------|------------|-------|
| **SEC-1** — Unbounded derivation-path DoS | **Yes** | Bounds enforced before `Vec` allocation and before HMAC loop |
| **SEC-2** — Mnemonic not zeroized on parse failure | **Yes** | `Zeroizing<String>` covers all exit paths |
| **SEC-3** — `Clone` on move-only sensitive wrappers | **Partial** | Core wrappers fixed; related types still clone secrets |

---

### SEC-1 — Derivation path DoS

**What the report asked for:** Reject oversized/deep paths at the API boundary before full parse and per-child HMAC work.

**What the PR does well:**

- `check_path_bounds` is cheap (length + slash count), runs inside `DerivationPath::from_str` before `collect`, and again via `check_derivation_path` on public entrypoints — good defense in depth.
- `generate_wormhole_from_seed` now calls `check_derivation_path` **before** the wormhole chain-id check, so junk paths fail fast instead of doing extra splitting first.
- Explicit `PathTooLong` / `PathTooDeep` errors map cleanly to `HDLatticeError`.
- Tests cover oversize byte length, over-depth, and the exact max-depth boundary.

**Limits vs real usage:** Documented paths like `m/44'/189189'/0'/0'/0'` (5 slashes) and wormhole paths like `m/44'/189189189'/0'` (3 slashes) sit well under `MAX_DERIVATION_DEPTH = 16` and `MAX_DERIVATION_PATH_BYTES = 256`.

**Minor nits (non-blocking):**

1. **Error wording:** `PathTooDeep` reports “segments” but the check counts `/` characters, not parsed `ChildNumber`s. For normal paths they match; odd paths like `m//1'` could differ. Consider renaming to “slashes” or counting parsed children after a cheap split if you want precision.

2. **Double parse on success:** `check_derivation_path` parses once; `ExtendedPrivKey::derive` parses again via `path.into()`. Cost is now bounded (≤16 HMAC steps), so this is an efficiency nit, not a security gap.

3. **`pub mod hderive`:** External callers can still call `ExtendedPrivKey::derive` directly, but `&str` paths still go through `from_str` → `check_path_bounds`. That’s acceptable.

---

### SEC-2 — Mnemonic zeroization on parse failure

**What the report asked for:** Ensure the owned mnemonic `String` is scrubbed on parse error, not only on success.

**Fix is correct:**

```90:100:hdwallet/src/lib.rs
pub fn mnemonic_to_seed(
	mnemonic: String,
	passphrase: Option<&str>,
) -> Result<[u8; 64], HDLatticeError> {
	// Drop guard: zeroizes the mnemonic on every exit path (success, parse error, unwind).
	let mnemonic = Zeroizing::new(mnemonic);
	let parsed_mnemonic = Mnemonic::parse_in_normalized(Language::English, mnemonic.as_str())
		.map_err(|e| HDLatticeError::Bip39Error(e.to_string()))?;
	Ok(parsed_mnemonic.to_seed_normalized(passphrase.unwrap_or("")))
}
```

Enabling `zeroize`’s `alloc` feature for `String: Zeroize` is the right dependency change.

**Test gap:** `test_mnemonic_to_seed_zeroizes_on_parse_failure` only asserts `Bip39Error`. It does not observe memory (hard in safe Rust). That’s typical for this class of fix; optional follow-up would be a short comment in the test that the guarantee is drop-guard semantics, not observable scrubbing.

**Pre-existing (out of PR scope):** `derive_key_from_mnemonic` still does `mnemonic.to_string()` before `mnemonic_to_seed`, so an extra heap copy exists before the `Zeroizing` wrapper. SEC-2’s reported bug is fixed; this is a separate hardening opportunity.

---

### SEC-3 — `Clone` vs documented move-only semantics

**What the report asked for:** Remove `Clone` from `SensitiveBytes32`, `SensitiveBytes64`, and `ExtendedPrivKey` so the API matches “move-only / no accidental copy.”

**Fixed as described:**

- `Clone` removed from both `SensitiveBytes*` types; docs updated.
- `ExtendedPrivKey` no longer derives `Clone`.
- Broken test updated to use two fresh wrappers from cloned **stack** seed bytes (reasonable for a determinism test).

**Still open (PR correctly flags this):**

- `Keypair` / `SecretKey` still derive `Clone` (`ml_dsa_87.rs`).
- `WormholePair` still derives `Clone` and stores raw `[u8; 32]` secrets.
- `SensitiveBytes32::into_bytes` / `as_bytes` and `SecretKey::to_bytes` still allow explicit copies — intentional, but worth documenting as distinct from “no `Clone`.”

**Docs drift:** Root [README.md](https://github.com/Quantus-Network/qp-rusty-crystals/blob/master/README.md) still claims compile-time copy prevention for `SensitiveBytes*` broadly; SEC-3 cited this. The PR updates struct docs but not the README. After merge, either update README or narrow the claim to the wrapper types only.

**Breaking change:** Removing `Clone` on public `SensitiveBytes*` types is a semver-breaking change for `qp-rusty-crystals-dilithium`. Worth a changelog / minor version bump note in the release.

---

### Code quality & tests

- DRY: single `check_path_bounds`, two call sites — good.
- Fail-early: no silent truncation — good.
- New tests are meaningful; boundary test (`test_derive_accepts_max_depth`) is especially useful.
- `test_fixed_seed_keypair` change is correct given move-only seeds.

---

### Recommendation

**Approve and merge** for the three reported issues. The implementation matches the audit intent.

**Before or right after merge (short follow-ups):**

1. Sync root README security claims with post-PR reality (and remaining `Clone` on `Keypair` / `WormholePair`).
2. Plan the flagged follow-up: drop `Clone` from `Keypair`, `SecretKey`, `WormholePair` when you can take an API bump.
3. Optional: tighten `derive_key_from_mnemonic` to avoid the extra `to_string()` copy; clarify `PathTooDeep` metric in the error message.

I did not re-run the full workspace test suite or clippy here; the PR description says both are green. If you want, I can run `./run_all_tests.sh` and `cargo clippy --workspace --all-targets -- -D warnings` locally before you merge.

n13:
## Re-review after follow-ups 1 & 2

Checked the current tree against the two items from the last review. Workspace tests and `cargo clippy --workspace --all-targets -- -D warnings` both pass.

---

### 1. README sync — **Done well**

The Security Features section now matches the implementation:

- Lists every move-only secret type (`SensitiveBytes32/64`, `Keypair`, `SecretKey`, `ExtendedPrivKey`, `WormholePair`).
- Documents explicit opt-in duplication via `to_bytes()` / `from_bytes()` with a concrete example.
- Calls out `Zeroizing<String>` on all `mnemonic_to_seed` exit paths.
- Documents bounded derivation paths with the actual constants.

This closes the docs drift from SEC-3 and the original audit README complaint. No overstated “compile-time prevents all copying” claim anymore.

**Tiny nuance (non-blocking):** `WormholePair` has public `[u8; 32]` fields, so a plain assignment `let s = pair.secret` still copies 32 bytes without `Clone`. The wormhole module doc already steers toward `&WormholePair`; you could add one line in the root README that wormhole secrets are plain arrays, not `to_bytes()` wrappers — optional polish only.

---

### 2. `Clone` removal on `Keypair`, `SecretKey`, `WormholePair` — **Done well**

| Type | `Clone` removed | Call-site migration |
|------|-----------------|---------------------|
| `Keypair` | Yes | `stack_usage_demo.rs` uses `to_bytes()` + `from_bytes()` in closures |
| `SecretKey` | Yes | — |
| `WormholePair` | Yes | Tests already used fresh `SensitiveBytes32` / references |

- Struct-level docs explain the move-only contract and the explicit copy path.
- No remaining `keypair.clone()` / `WormholePair` clone usage in the repo.
- SEC-3 is now **fully addressed** for the types named in the audit and in your README.

**Release note:** This is a **breaking API change** for `qp-rusty-crystals-dilithium` and `qp-rusty-crystals-hdwallet` consumers that relied on `Keypair`/`SecretKey`/`WormholePair: Clone`. Worth a minor/major bump and changelog entry when you ship.

---

### Original audit items (unchanged, still good)

- **SEC-1** — Path bounds before alloc/HMAC; tests still valid.
- **SEC-2** — `Zeroizing<String>` on mnemonic; parse-failure path covered.
- **SEC-3** — Complete with this round of `Clone` removals.

---

### Still optional (not part of 1 & 2)

These were never claimed fixed; leaving them here for completeness:

- **`derive_key_from_mnemonic` / `derive_wormhole_from_mnemonic`** still do `mnemonic.to_string()` before `mnemonic_to_seed`, so an extra heap copy exists before the zeroizing wrapper.
- **`PathTooDeep`** error text still says “segments” while `check_path_bounds` counts `/` characters (behavior is fine for normal paths; wording could be tightened later).

---

### Verdict

**Approve.** Follow-ups 1 and 2 are implemented correctly; the PR is stronger than the first review and the three audit findings are covered end-to-end. Remaining nits are optional hardening, not blockers for merge.

n13:
Opus 4.7 max review 

Build clean, clippy clean with `-D warnings`, and all four new tests pass on `fix_security_audit_issues` ([8d69df8..f9a9a11](https://github.com/Quantus-Network/qp-rusty-crystals/pull/79)). Here's my fresh independent review.

---

## TL;DR

The three audit findings are addressed correctly and the implementation is solid. I'd **approve and merge**. Below is what I verified, what I think is well done, and the small set of residual nits worth tracking — most of which the author already flagged in their self-reviews.

---

## Verification I ran locally

- `cargo check --workspace --all-targets` → clean.
- `cargo clippy --workspace --all-targets -- -D warnings` → clean.
- `cargo test -p qp-rusty-crystals-hdwallet --lib -- test_derive_rejects_oversized_path test_derive_rejects_too_deep_path test_mnemonic_to_seed_zeroizes_on_parse_failure test_derive_accepts_max_depth` → 4/4 pass.
- Searched the repo for remaining `.clone()` on the just-de-`Clone`d types (`SensitiveBytes32/64`, `Keypair`, `SecretKey`, `WormholePair`, `ExtendedPrivKey`). None found. The `.clone()` calls in `ct_bench.rs` are all on non-secret polynomial / nonce / message types — unrelated.

So the author's claim that everything builds and tests cleanly checks out.

---

## SEC-1 — Unbounded derivation-path DoS

**Verdict: fully addressed.**

Defense-in-depth structure is good:

```19:30:hdwallet/src/hderive.rs
/// Reject paths whose raw byte length or `/`-segment count exceed the workspace caps.
/// Runs before any allocation so attacker-controlled paths cannot drive memory or CPU.
pub(crate) fn check_path_bounds(path: &str) -> Result<(), Error> {
	if path.len() > MAX_DERIVATION_PATH_BYTES {
		return Err(Error::PathTooLong(path.len()));
	}
	let depth = path.bytes().filter(|b| *b == b'/').count();
	if depth > MAX_DERIVATION_DEPTH {
		return Err(Error::PathTooDeep(depth));
	}
	Ok(())
}
```

- `check_path_bounds` runs **before** `Vec<ChildNumber>` is collected in `DerivationPath::from_str`, and again at the public entrypoint via `check_derivation_path` → cheap, runs first, no allocation/HMAC executed on hostile input.
- DRY: single source of truth, two call sites, exactly the structure the user rules ask for.
- Errors are explicit (`PathTooLong(usize)` / `PathTooDeep(usize)`) and propagate cleanly to top-level `HDLatticeError`. Fail-early, no silent truncation — matches the user rule.
- `generate_wormhole_from_seed` now runs `check_derivation_path` **before** the wormhole `nth(2)` check ([`hdwallet/src/lib.rs` L155](https://github.com/Quantus-Network/qp-rusty-crystals/pull/79/files)). This is the right order, since the chain-id check operates on a lazy `split` iterator and `nth(2)` would *not* allocate on a long path, but failing bounds first keeps the path-length check single-source-of-truth.
- Constants are well-chosen: 16 segments / 256 bytes leaves the legitimate `m/44'/189189'/0'/0'/0'` (5 segs) and wormhole `m/44'/189189189'/0'` (3 segs) deep inside the budget.

The boundary test (`test_derive_accepts_max_depth`) is the right shape — exactly-16 segments must still derive successfully — and confirms the off-by-one is correct.

**Minor nit (non-blocking, already flagged by the author):** `PathTooDeep` reports "segments" but `check_path_bounds` counts `/` characters. For well-formed paths these are equal; for pathological inputs like `m//1'` they differ. The wording is fine in practice, but renaming the constant to `MAX_DERIVATION_SLASHES` or counting parsed children after split would be more precise.

**Efficiency nit (out of scope):** `check_derivation_path` parses once; `ExtendedPrivKey::derive` reparses via `path.into()` → `from_str`. Cost is now bounded to ≤16 segments, so this is just a couple of redundant `Vec` allocations, not a security gap. Could be deduplicated by passing the already-parsed `DerivationPath` through, but it would tangle the public API.

---

## SEC-2 — Mnemonic not zeroized on parse failure

**Verdict: fully addressed.**

```92:101:hdwallet/src/lib.rs
pub fn mnemonic_to_seed(
	mnemonic: String,
	passphrase: Option<&str>,
) -> Result<[u8; 64], HDLatticeError> {
	// Drop guard: zeroizes the mnemonic on every exit path (success, parse error, unwind).
	let mnemonic = Zeroizing::new(mnemonic);
	let parsed_mnemonic = Mnemonic::parse_in_normalized(Language::English, mnemonic.as_str())
		.map_err(|e| HDLatticeError::Bip39Error(e.to_string()))?;
	Ok(parsed_mnemonic.to_seed_normalized(passphrase.unwrap_or("")))
}
```

The fix is the right shape: by wrapping into `Zeroizing<String>` **before** the fallible `parse_in_normalized`, the `Drop` impl guarantees scrubbing on every exit — success, parse error via `?`, and panic-unwind during PBKDF2. The corresponding `zeroize = { workspace = true, features = ["alloc"] }` change in `hdwallet/Cargo.toml` brings in the `String: Zeroize` impl. Correct.

The test only asserts the error variant — it does not, and in safe Rust cannot easily, observe the buffer post-drop. That's acceptable for this class of fix; the guarantee is structural (`Zeroizing` drop-guard), not observational.

**Pre-existing, out of scope:** `derive_key_from_mnemonic` / `derive_wormhole_from_mnemonic` do `mnemonic.to_string()` on the `&str` argument before calling `mnemonic_to_seed`. The cloned `String` is zeroized correctly, but the caller's original `&str` buffer is not under the function's control. SEC-2 as reported is fixed; tightening these wrappers is a worthwhile follow-up but not a blocker.

---

## SEC-3 — `Clone` on move-only sensitive types

**Verdict: addressed, including the audit's named types plus the follow-up `Keypair` / `SecretKey` / `WormholePair`.**

- `Clone` removed from `SensitiveBytes32`, `SensitiveBytes64`, `ExtendedPrivKey`, `Keypair`, `SecretKey`, `WormholePair`. Doc comments on each call out that this is deliberate.
- The only call site that broke (`test_fixed_seed_keypair`) now constructs two fresh `SensitiveBytes32` from the same fixed `[0x55u8; SEEDBYTES]`, which is the test's actual intent. Cleaner than the previous `seed.clone()`.
- `stack_usage_demo.rs` migrated to `Keypair::to_bytes()` + `from_bytes()`, which is the documented escape hatch — good demonstration of the intended pattern.

**Substantive nit worth flagging (the only one I'd actually act on):** The README and the new doc-comment on `WormholePair` claim "move-only" / "compile-time copy prevention," but `WormholePair` exposes its secret as a public field:

```47:55:hdwallet/src/wormhole.rs
#[derive(Eq, PartialEq, ZeroizeOnDrop)]
pub struct WormholePair {
	/// Deterministic Poseidon-derived address.
	pub address: [u8; 32],
	/// First hash of secret
	pub first_hash: [u8; 32],
	/// The hashed secret used to generate this address.
	pub secret: [u8; 32],
}
```

Because `[u8; 32]` is `Copy`, `let s = pair.secret;` still produces a fresh, non-zeroized copy of the secret bytes on the stack — no `Clone` required. Removing `derive(Clone)` blocks `wormhole_pair.clone()` but not field access. Two reasonable options for a follow-up:

1. Make the fields private and add `secret()` returning `&[u8; 32]` (or wrap in `SensitiveBytes32`) so the escape hatch is explicit like everywhere else.
2. Keep the public fields but soften the README claim for `WormholePair` to "no derived `Clone`; secret remains a public `[u8; 32]` for ergonomic access — prefer `&WormholePair`."

The current README points users at `to_bytes()`/`from_bytes()` as the explicit copy idiom, but `WormholePair` doesn't actually have those methods — only field reads. That's a small docs/reality drift worth one line of cleanup.

**Release-engineering reminder:** Removing `Clone` on public `Keypair`, `SecretKey`, `SensitiveBytes32/64`, `ExtendedPrivKey`, `WormholePair` is a semver-breaking change for both crates. Worth a minor/major version bump and a one-line changelog when this ships.

---

## Code quality observations

- DRY-clean: `check_path_bounds` is the single source of bounds logic; `check_derivation_path` is the single source of error mapping. Matches the project rules.
- Fail-early: no fallbacks, errors are explicit and typed. Matches the project rules.
- Comments are minimal and explain intent (the move-only contract, the defense-in-depth ordering) — not narration. Matches the project rules.
- `pub(crate) fn check_path_bounds`: correct visibility — external callers go through `from_str`, which already invokes it.
- `DerivationPath::path` stays private, so `IntoDerivationPath for DerivationPath` cannot be used to bypass bounds (you can only get an oversized `DerivationPath` by going through `from_str`, which already enforces bounds). Safety is closed.

---

## Recommendation

**Approve.** All three audit findings are correctly fixed, the implementation is tight, and the follow-ups go beyond the audit scope in a useful way. The remaining nits are:

1. (Worth doing) Tighten the README / `WormholePair` story so the public-field copy path is acknowledged, or hide the field behind an accessor.
2. (Optional, future PR) Either skip the redundant `to_string()` in `derive_key_from_mnemonic`/`derive_wormhole_from_mnemonic` or document that the caller's mnemonic is not scrubbed there.
3. (Optional cosmetic) Rename / reword `PathTooDeep` so "segments" matches the metric being counted (slashes).
4. (Release engineering) Version bump + changelog for the breaking `Clone` removal.

None of those block the merge. The PR makes the codebase materially safer against the three reported issues without introducing new sharp edges.

I have not commented on GitHub (per the do-not-auto-comment rule). Happy to push any of the above as follow-up commits on this branch, or open issues for the post-merge tracking — just say the word.

illuzen (approved):

PR #80, opened 2026-06-03
add hardening tests

n13:
## Summary

Bernstein's *"Exploiting ML-DSA bugs"* (2026-06-01) describes three implementation
bug classes that enable secret-key recovery from 1–2 signatures. An audit found this
implementation **unaffected**; this PR adds focused regression tests that lock in that
safe behavior, plus a small robustness guard. No signing/keygen behavior changes — the
NIST KAT remains byte-exact.

**Regression tests (one per bug class):**
- **Class 1 – AABBCC y-coefficient repetition** — `test_uniform_gamma1_no_paired_coefficient_repetition` (`poly.rs`): over 256 seeds, asserts the `coeffs[2i] == coeffs[2i+1]` collision rate stays near the ~1/2²⁰ chance level (a repetition bug forces it to ≥ ½), plus an all-even/all-odd-zero check for A0B0 patterns.
- **Class 2 – K zeroed/omitted** — `test_secret_key_k_affects_signature` (`sign.rs`): flips one byte of the stored `K` and asserts the signature changes (and still verifies), proving `K` actually feeds the mask seed.
- **Class 3 – repeated y nonce** — `test_y_differs_across_messages_deterministic` (`sign.rs`): signs two messages deterministically (`hedge=None`) under one key, recovers the actual `y = z − c·s1` from each signature, and asserts the masks differ. `test_mask_seed_golden_vector` (`sign.rs`) additionally pins `ρ' = H(K‖rnd‖μ)` to an independent Python `shake_256` vector and cross-checks the incremental-absorb path against one-shot SHAKE256.

**Supporting changes (`sign.rs`):**
- Split `prepare_signing_context` into `derive_message_hash` (μ) and `derive_mask_seed` (ρ') helpers (DRY; lets the golden test pin the real production path). Behavior-preserving.
- Added an `assert!` guarding `attempt_nonce` against u16 overflow of the `L*nonce + i` mask nonce — fails loudly instead of silently reusing a nonce in the astronomically unlikely many-restart case.

## Test plan
- [x] `cargo test -p qp-rusty-crystals-dilithium` — 87 passed (+4 new)
- [x] NIST KAT byte-exact: `test_nist_kat` ok (confirms the refactor changes no output)
- [x] `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` — clean

## Reference PDF "Exploiting ML-DSA bugs"

[mldsa-20260601.pdf](https://github.com/user-attachments/files/28537586/mldsa-20260601.pdf)


illuzen (approved):
LGTM

PR #86, opened 2026-06-08
threshold: bind combine_with_message to round 2 message; fix max-parties doc

n13:
## Summary

Follow-up to the re-review of #74, addressing the two actionable residual items called out there. Based on `illuzen/threshold-2` so it can be folded into #74 (the `threshold` crate only exists on that branch).

- **`combine_with_message` now binds to the Round 2 message/context.** Previously it used the caller-supplied `message`/`context` and ignored the values captured in `AfterRound3`; combining against a different message silently produced an invalid signature (`CombinationFailed`) rather than failing on the real cause. It now rejects a mismatch with `InvalidData` and uses the bound values for the actual signature computation. (`combine()` already used the captured values, and the network adapter passes its own stored message — so this only tightens the standalone convenience API.)
- **Doc fix:** `lib.rs` claimed support "up to (7, 7)" while `MAX_PARTIES = 6`.

Not included (by design / not closable in a patch — these need an external cryptographic review and are already covered by the crate's "research only — not audited" warning): no in-library CSPRNG / SSID not binding the message, floating-point hyperball sampling and non-constant-time NTT on secret data, and the hardcoded hyperball/`K` parameters resting on a Monte-Carlo script rather than a published proof.

## Test plan

- [x] `cargo test -p qp-rusty-crystals-threshold --lib` (201 passed)
- [x] `cargo test -p qp-rusty-crystals-threshold --test signer_state_tests` (13 passed, incl. new `test_combine_with_message_rejects_mismatched_message`)
- [x] `cargo clippy -p qp-rusty-crystals-threshold --all-targets` (clean)
- [x] `cargo +nightly fmt -p qp-rusty-crystals-threshold -- --check` (clean)

### DIFF
--- a/threshold/src/lib.rs
+++ b/threshold/src/lib.rs
@@ -11,7 +11,7 @@
 //! - Any t or more parties can cooperate to produce a valid signature
 //! - Fewer than t parties cannot produce a signature or learn the secret key
 //!
-//! This implementation supports configurations up to (7, 7) and produces
+//! This implementation supports configurations up to (6, 6) and produces
 //! signatures that are compatible with standard ML-DSA-87 verification.
 //!
 //! ## Quick Start
--- a/threshold/src/signer.rs
+++ b/threshold/src/signer.rs
@@ -687,15 +687,26 @@ impl ThresholdSigner {
 		_all_round2: &[Round2Broadcast],
 		all_round3: &[Round3Broadcast],
 	) -> ThresholdResult<Signature> {
-		let (round2_data, my_responses, _, _) = self.state.expect_round3()?;
+		let (round2_data, my_responses, bound_message, bound_context) =
+			self.state.expect_round3()?;
+
+		// Round 3 responses are bound to the message/context captured in Round 2.
+		// Combining against anything else would silently yield an invalid signature,
+		// so reject the mismatch instead of proceeding.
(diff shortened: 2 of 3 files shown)


n13:
## Finding
`threshold/tests/signer_state_tests.rs` adds a regression test, but the key assertion is too weak:

```319:321:threshold/tests/signer_state_tests.rs
let mismatched = signers[0].combine_with_message(b"different message", context, &r2, &r3);
assert!(mismatched.is_err(), "mismatched message must be rejected");
```

The old behavior also errored here, usually as `CombinationFailed`, so this test would not prove the fix. It should assert the specific `ThresholdError::InvalidData(...)` mismatch error, and ideally also cover mismatched context.

## Verdict
Yes, the code change itself improves things. It makes `combine_with_message` fail early with the real cause when the caller passes a message/context different from Round 2, instead of proceeding into signature combination and surfacing a misleading failure.

It is not strictly needed for the network adapter path, since that already passes stored message/context, but it is useful for the standalone API and matches the fail-early behavior we want.

The `(7, 7)` to `(6, 6)` doc fix is clearly correct based on `MAX_PARTIES = 6`.

I ran `cargo test -p qp-rusty-crystals-threshold --test signer_state_tests`: 13 passed. My recommendation: keep the PR, but tighten the test before merging.

PR #88, opened 2026-06-09
Add release workflows for threshold

illuzen:
> [!NOTE]
> **Medium Risk**
> Merging labeled release PRs can auto-tag, publish to crates.io, and push releases using privileged tokens; mistakes in version extraction or workflow inputs could ship the wrong artifact.
> 
> **Overview**
> Adds **automated threshold release CI** modeled on the existing dilithium/hdwallet flows: a manual **Create Threshold Release Proposal** workflow and a **tag & publish** workflow on merged PRs.
> 
> The proposal workflow computes the next `threshold-v*.*.*` tag (or a custom `v…` version), bumps `threshold/Cargo.toml`, the workspace `qp-rusty-crystals-threshold` pin in root `Cargo.toml`, and `Cargo.lock`, then opens a labeled PR via `ADMIN_PAT`. The follow-up workflow runs when a `threshold-release-proposal` PR merges to `master`: it tags from the PR title, runs format/build/test/clippy/docs and `cargo audit` on that tag, then creates a GitHub release (optional draft) and **`cargo publish`**es `qp-rusty-crystals-threshold` with `CARGO_REGISTRY_TOKEN`.
> 
> **`threshold/Cargo.toml`** metadata is updated for crates.io (Quantus authors, ML-DSA-87 description, readme/homepage/repo/keywords)—no dependency or version field change in this diff.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17973b5a261395b7de4488dff8552e70ab824207. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/create-threshold-release-proposal.yml
+++ b/.github/workflows/create-threshold-release-proposal.yml
@@ -62,9 +62,18 @@ jobs:
           # Get all threshold version tags and sort them by version
           latest_semver_tag=$(git tag -l "threshold-v[0-9]*.[0-9]*.[0-9]*" | sort -V | tail -n 1)
           
-          # If no tags found, use default
+          # If no tags found, read current version from threshold/Cargo.toml
           if [ -z "$latest_semver_tag" ]; then
-            latest_semver_tag="threshold-v0.0.0"
+            cargo_version=$(grep -m1 '^version' threshold/Cargo.toml | sed 's/version = "\(.*\)"/\1/')
+            if [ -z "$cargo_version" ]; then
+              echo "Error: No threshold tags found and could not read version from threshold/Cargo.toml"
+              exit 1
+            fi
+            # Use the Cargo.toml version as the base (subtract one patch for the first bump to match)
+            # This ensures a patch bump from Cargo.toml 1.0.0 yields 1.0.1, not 0.0.1
+            latest_semver_tag="threshold-v$cargo_version"
+            echo "No existing threshold tags found. Using version from threshold/Cargo.toml: $latest_semver_tag"
+            echo "Note: The first release will increment from this version."
           fi
           
           echo "latest_tag_found=$latest_semver_tag" >> $GITHUB_OUTPUT
@@ -171,7 +180,7 @@ jobs:
 
           gh pr create \
             --title "$PR_TITLE" \
-            --body "$(printf "Automated threshold version bump for release %s.\\n\\n%s\\n\\nTriggered by workflow run: %s\\n\\nType: %s" "$NEW_VERSION" "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" "${{ github.event.inputs.version_type }}")" \
+            --body "$(printf "Automated threshold version bump for release %s.\\n\\nTriggered by workflow run: %s\\n\\nType: %s" "$NEW_VERSION" "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" "${{ github.event.inputs.version_type }}")" \
             --base "$TARGET_BRANCH" \
             --head "$branch_name" \
             --label "$PR_LABELS"
--- a/.github/workflows/create-threshold-release-proposal.yml
+++ b/.github/workflows/create-threshold-release-proposal.yml
@@ -1,0 +1,177 @@
+name: Create Threshold Release Proposal
+
+env:
+  CARGO_TERM_COLOR: always
+
+on:
+  workflow_dispatch:
+    inputs:
+      target_branch:
+        description: 'Target branch for the PR (default: master)'
+        required: false
+        type: string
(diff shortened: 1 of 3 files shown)


n13:
I've reviewed the current state of PR #88 thoroughly. Here's my assessment.

## Verdict: One blocking bug introduced by the CodeQL autofix; everything else is resolved

### Blocker — the publish workflow is now broken and will fail on every merge to `master`

Commit `7dd8881` (the "Checkout of untrusted code" CodeQL autofix) changed the trigger of `create-threshold-release-tag-and-publish.yml`:

```7:20:.github/workflows/create-threshold-release-tag-and-publish.yml
on:
  push:
    branches:
      - master

# Default to read-only; jobs that need to push tags or create releases
# override this with explicit `contents: write` (see create-tag, create-github-release).
permissions:
  contents: read

jobs:
  create-tag:
    name: Create Threshold Tag
    if: github.ref == 'refs/heads/master'
```

But the rest of the job still reads the **pull request** context, which does not exist on a `push` event:

```35:53:.github/workflows/create-threshold-release-tag-and-publish.yml
        env:
          PR_TITLE: ${{ github.event.pull_request.title }}
          PR_LABELS: ${{ join(github.event.pull_request.labels.*.name, ',') }}
        run: |
          VERSION=$(echo "$PR_TITLE" | grep -o 'threshold-v[0-9]\+\.[0-9]\+\.[0-9]\+')
          if [ -z "$VERSION" ]; then
            echo "Error: Could not extract threshold version from PR title: $PR_TITLE"
            exit 1
          fi
```

Consequences:
1. On a `push` event `github.event.pull_request.title` and `.labels` are always empty, so the version `grep` matches nothing and the job **exits 1 on every push to `master`** — noisy failing runs on every merge.
2. The release mechanism can never work: no version is extracted, so no tag is created, nothing is published, and draft detection always reports non-draft.
3. `if: github.ref == 'refs/heads/master'` now fires for *every* merge, not just release-proposal PRs.

The sibling `dilithium`/`hdwallet` publish workflows use the correct pattern, and threshold should match it:

```7:21:.github/workflows/create-hdwallet-release-tag-and-publish.yml
on:
  pull_request:
    types: [closed]
    branches:
      - master
...
    if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'hdwallet-release-proposal')
```

**Recommendation:** revert the trigger/condition back to `pull_request: types: [closed]` + the `merged == true && contains(..., 'threshold-release-proposal')` guard. The CodeQL "Checkout of untrusted code in a trusted context" alert is a false positive here (the release-proposal branch is repo-internal and gated by a label only the automation sets) — it should be dismissed the same way it must already be handled for dilithium/hdwallet, **not** "fixed" by switching to `push`. Note the CI shows CodeQL green only because `push` sidesteps the analysis, masking that the workflow is non-functional.

### Resolved since the last review (verified in the current branch)
- **printf placeholder mismatch** (proposal body): now 3 `%s` / 3 args. ✅
- **First-bump downgrade** (no tags → `v0.0.0`): now reads the version from `threshold/Cargo.toml`. ✅
- **Draft still publishes**: `cargo publish` is now gated by `if: needs.create-tag.outputs.is_draft != 'true'`. ✅
- **Code injection** (PR title/actor): now passed via env vars. ✅
- `threshold/Cargo.toml` metadata is correct, and the declared `readme = "README.md"` exists, so `cargo publish` won't fail on a missing readme. ✅

### Minor nits (non-blocking, mostly inherited from the existing workflows)
- Misleading comment in the proposal workflow: it says it will "subtract one patch... so 1.0.0 yields 1.0.1," but the code does no subtraction. With `Cargo.toml` at `1.0.0`, the first patch release is `1.0.1` — version `1.0.0` is never published. Worth aligning the comment with the behavior (or deciding `1.0.0` should be the first tag).

```72:76:.github/workflows/create-threshold-release-proposal.yml
            # Use the Cargo.toml version as the base (subtract one patch for the first bump to match)
            # This ensures a patch bump from Cargo.toml 1.0.0 yields 1.0.1, not 0.0.1
            latest_semver_tag="threshold-v$cargo_version"
```

- The root `Cargo.toml` pin is `version = "=1.0.0"`, but the proposal `sed` replaces the whole quoted value with a bare `1.0.1`, dropping the `=` exact-pin. This is the same pattern as dilithium/hdwallet, so it's pre-existing, but it's inconsistent with the exact-pin convention used throughout the workspace.
- DRY: these three release workflow pairs are ~95% identical. Per the repo's DRY preference, a single parameterized reusable workflow (crate name/prefix as inputs) would remove a lot of duplication — out of scope for this PR, but worth a follow-up.

Want me to apply the trigger/condition fix (revert to `pull_request: types: [closed]` + the merged/label guard) on the `illuzen/threshold-ci` branch, or just leave the review here?

PR #89, opened 2026-06-09
Resharing hardening

illuzen:
- add check for recovered partial share norm
- have old committee members check dealer commitment
- add sketch for security proof
- update README with more accurate numbers

---

> [!NOTE]
> **Medium Risk**
> Changes verification on the committee handoff path for threshold signing keys; abort behavior and accepted-share criteria change, though the public key and hyperball parameters are unchanged.
> 
> **Overview**
> **Hardens the threshold ML-DSA resharing protocol** with two runtime checks and supporting documentation.
> 
> After Round 3 completes, **old-committee peers** who share an RSS subset but are not the designated dealer now **recompute expected sub-shares** and **abort if dealer commitments mismatch**, blocking bad Round 3 hashes before Round 4 private delivery.
> 
> In Round 5, **new committee members** run **`verify_recovered_partial_norms`**: for every threshold signing set they belong to, they recover the same partial secret signing would use and reject outputs where **τ × weighted norm exceeds the existing hyperball `r'`** for `(t_new, n_new)`. This targets **public-key-preserving zero-sum reshaping** that can pass per-coefficient bounds but break the signing proof envelope.
> 
> Adds **`SECURITY_PROOF.md`** (proof outline, invariants, reduction sketch), refreshes **`resharing/README.md`** (protocol steps, guard margins, honest vs adversarial framing), a **Deep Wiki** badge in the root **`README.md`**, and **unit tests** for peer commitment verification and the norm guard.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14208bd2a9f7e242e2690f302dfcd0c12ca7b503. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13:
Claude Fable 5 max review

# Review: PR #89 — Resharing hardening

**Verdict: solid hardening, sound implementation, recommend approve with minor follow-ups.** CI is green, all 4 commits build, and the full resharing suite plus the two new unit tests pass locally. I verified the crypto-relevant details below.

## What I verified as correct

**Round 3 peer check (`verify_peer_dealer_commitments`)**

- Recomputation is byte-exact with the dealer path: same `derive_subshares_with_session_seed`, same `commit_subshare`, same `designated_dealer_for` (lowest set bit), same `new_subset_order`. All members of an old subset hold identical `s_I` (guaranteed by keygen and the prior resharing's Round 5 consistency check), so honest dealers can never trip it.
- It runs exactly at the `Round3Waiting → Round4Generate` transition, so a detecting peer halts before sending its own Round 4 shares. `have_enough_round3()` requires all old participants, so the dealer's broadcast is always present when the check runs — the `DealerDeliveryFailed` lookups can't spuriously fire.
- The `types.rs` doc fix is a real correction: the old comment claimed peers verify dealer commitments "in Round 5", which was never implemented on master.

**Round 5 norm guard (`verify_recovered_partial_norms`)**

- The permutation and pattern-translation logic exactly mirrors `recover_share` in `secret_sharing.rs` (sorted active indices, active-first/inactive-after perm, identical mask translation), and since `perm[current_i] = my_idx`, every translated mask is a subset this party actually holds — no false "missing share" errors.
- Signing enforces **exactly t** participants (`signing_protocol.rs` line 599), so enumerating size-t signing sets covers every partial this party can ever use. Complete coverage.
- The weighting `sqrt(‖s1‖²/ν² + ‖s2‖²)` matches `HyperballSampleVector::excess`, and `‖c·s‖₂ ≤ τ·‖s‖₂` is a valid bound for negacyclic multiplication by a τ-sparse ±1 challenge.
- A subtle point that works out in the right direction: the guard sums *centered* coefficients in `i64` without mod-Q reduction, while signing reduces mod Q and then centers. Since `|centered(x)| ≤ |x|`, the guard computes an **upper bound** on the true signing norm — conservative, never under-estimates. Honest sums (~2500 max) can't wrap Q/2 ≈ 4.2M, so honest executions are unaffected. f64 precision is negligible at the decision boundary.
- Failure placement is graceful: `success=false` in the Round 5 broadcast → all parties abort in `Combining`. The README's updated margin tables are arithmetically correct (e.g. 60×2894 = 173,640; 1 − 173640/517853 = 66.5%).
- The new unit test genuinely demonstrates the closed gap: ±450 per-coefficient zero-sum noise passes `SUBSHARE_COEFF_BOUND = 500` and preserves the PK, but is caught only by the norm guard.

Also: GitHub's "hidden or bidirectional Unicode" banner is a false alarm — I scanned all changed files; there are no bidi/invisible characters, only UTF-8 box-drawing and arrow characters in docs/comments.

## Findings

**1. The guard radius `r'` is generous, and the proof sketch concedes it (main question for the author).**
Using the full sampling radius as `R_guard` leaves config-dependent slack over honest behavior: ~3x for 4-of-6 but **~59x for 2-of-3** (honest max τ·norm 10,740 vs r' = 631,703). An adversary can still inflate recovered partials substantially within acceptance. `SECURITY_PROOF.md` itself acknowledges this ("For a paper-style instantiation against the exact Mithril theorem, instantiate `R_guard` with a bound `B` satisfying the hyperball parameter-selection condition"). If the signing proof yields a concrete per-config `B`, the guard should enforce that instead of `r'` — or the docs should state why `r' `suffices. As-is it's a meaningful catastrophic-reshaping cap, but the README line "the same norm guard used to justify signing with the existing parameters" slightly overstates what's enforced.

**2. DRY: the guard re-implements `recover_share`'s recovery logic.**
The perm/translate block in `recovered_partial_weighted_norm` is a copy of the index logic in `secret_sharing.rs` — and the guard's entire value rests on it staying identical to signing. If `recover_share` ever changes, the guard silently diverges. Extracting the permutation+translation into a shared helper in `secret_sharing` would make agreement structural rather than coincidental. Same (smaller) note for the dealer-commitment lookup duplicated between `verify_peer_dealer_commitments` and `verify_and_aggregate_new_shares`.

**3. Abort asymmetry / liveness.**
Round 1–2 failures are detectable by *everyone* (public data), and Round 5 failures are *broadcast*. The new Round 3 peer check is the first failure only a subset of parties can observe, and it aborts via a local `Err` without notifying anyone — new-only members will hang until a transport timeout. Halting before sending one's own Round 4 shares is the right security call; just flagging that the surrounding runner must handle this via timeout, since the protocol itself won't converge to a global abort here.

**4. No e2e test exercises the norm-guard path.**
`test_resharing_detects_consistent_dealer_tamper_at_t_equals_n` covers PK-breaking tampering, but a *consistent* (Round 3 + Round 4) zero-sum tamper across two new subsets — which preserves the PK and passes coefficient bounds — is only covered by the unit test that injects `new_shares` directly. The existing `TamperFn` harness could exercise the full path including the graceful abort.

**Nits**

- `compute_sharing_patterns` is recomputed inside the per-signing-set loop; it's mask-independent and could be hoisted (micro, configs are tiny).
- README: "before any Round 4 private delivery occurs" is true for the verifying peer's own deliveries, but in an async network a malicious dealer's Round 4 sends aren't blocked. `SECURITY_PROOF.md` phrases this correctly; the README slightly over-promises.
- The DeepWiki badge commit is unrelated to the hardening, but harmless.

Per your rules I haven't posted anything to GitHub — let me know if you'd like me to submit any of this as a review there.

n13:
## Review: Resharing hardening

I read the full diff and cross-checked the security-critical Rust against the existing signing/recovery code. Overall this is a careful, defensible hardening change — both new guards are correct and fail-closed. Two things I'd want addressed before merge, plus items for an external crypto sign-off.

### Correctness — verified good
- **Round-5 guard recovery faithfully mirrors signing.** The `perm` construction + pattern→subset translation in `recovered_partial_weighted_norm` matches `recover_share` (secret_sharing.rs) exactly. Every translated mask is a size-`k_new` subset containing `my_idx`, which `verify_and_aggregate_new_shares` always inserts, so recovery is complete or aborts safely.
- **Coefficient-domain centering is sound.** `new_shares` are in `[0,Q)`; the guard sums `center_mod_q` in i64. With `SUBSHARE_COEFF_BOUND=500` and ≤10 subsets/pattern, max magnitude ~5000 ≪ Q/2 ≈ 4.19M, so sum-of-centered == centered-of-sum (no wraparound) — the guard measures the true partial-secret norm.
- **Fail-closed + abort wiring is correct.** Guard error → `success=false` → Round-5 broadcast → `handle_combining` aborts. Unsupported new config → `get_hyperball_params` returns `None` → error.
- **New noise splitter is exact + deterministic.** `add_mean_subtracted_noise` is integer zero-sum; `sample_split_noise_coeff` consumes exactly one byte; `split_noise_threshold` clamps to `[1,127]` so `2T ≤ 254 < 256`. Determinism across old-subset peers is preserved (which is what makes peer verification meaningful).
- **Parameter consistency checks out.** Radii are `κ·base` and `partial_secret_norm_bound` applies the same κ. Message limits are mutually consistent (`k=1600` ⇒ 9.42 MB ≤ `MAX_COMMITMENT_DATA_SIZE` 10.5 MB ≤ `MAX_SIGNING_MESSAGE_SIZE` 12 MB). `binomial` is exact/overflow-safe for n≤6.

### Medium — acceptance/κ path isn't validated in default CI
`analyze_recovered_partials` only asserts the trivial `max_combined_norm < r_prime` (~99% margin) and merely *prints* the real overshoot vs `B`. So no automated test asserts honest reshares stay within `κ·B` (the quantity the new guard enforces and the basis for κ selection). The honest-reshare-passes-guard tests (`test_reshares_and_sign_3_of_5/_4_of_6`), `test_recovered_partial_variance_4_of_6`, `test_coefficient_distribution_4_of_6`, and `test_dealer_sign_4_of_6` are all `#[ignore]`d. Net: **`(4,6)` — the config this PR exists to enable — has zero default-CI coverage**, and the only guard test that runs in CI is the *rejection* path.

Suggest: add an assertion in `analyze_recovered_partials` that overshoot ≤ the config's κ for the cheap configs that already run un-ignored (`(2,2)`/`(2,3)`/`(2,4)`/`(3,5)`), so the κ margins are regression-protected.

### Medium — DRY: duplicated signing-set recovery logic
`recovered_partial_weighted_norm` re-implements the `perm` build + pattern translation already in `recover_share`. Extracting a shared helper (e.g. `translated_subset_masks(signing_mask, my_idx, threshold, parties)`) with each caller doing its own NTT vs centered-i64 accumulation would prevent the two copies drifting apart — which would be a silent correctness bug given the guard's whole purpose is to match signing.

### Low / for the crypto auditor
- **`√τ` vs `τ`**: the guard uses the expected (Gaussian-heuristic) `√τ`, matching how `B` is defined (Mithril §3.4) and consistent with signing's rejection sampling. Acceptable, but leans on the threshold-ML-DSA proof using the same convention — worth explicit sign-off.
- **`(2,2)` key-hiding parity** is measured at 1.99 vs base 2.00, i.e. marginally *below* parity (claimed within MC error).
- **`(4,6)` cost**: K=1600 ⇒ ~15 MB/session, ~7s/sign release, `Q_s ≈ 2^28.2`. Confirm near-mpc transport timeouts/limits handle 10–12 MB messages.
- **Unaudited supporting math**: `SECURITY_PROOF.md` is honest that it's heuristic, not an audit; the numeric tables come from python scripts not run in CI. Consider wiring at least `keyhiding_conditional.py` into CI (or pinning output) so the hiding claim doesn't silently rot.

### Verdict
Core hardening is sound and the guards are correct and fail-closed. Before merge I'd want (1) an automated assertion that honest reshares pass the `κ·B` guard for un-ignored configs, and ideally (2) the DRY refactor of the recovery logic. The key-hiding/`√τ` reasoning is heuristic-but-measured and should get an external crypto reviewer's sign-off given this is committee-handoff code for signing keys.

n13:
should do (1) automated assertion

n13:
# Review: PR #89 — Resharing hardening

_Quantus-Network/qp-rusty-crystals · branch `illuzen/resharing-hardening`_

I read the full diff and traced the security-critical Rust against the existing signing/recovery code. Both new guards are correct and fail-closed, the DRY refactor the earlier reviews asked for has landed cleanly, and the doc overclaim (comparing `τ·‖p‖` against the sampling radius `r'`) is now correctly stated as `√τ·‖p‖_ν ≤ B'`.

**Verdict: sound engineering, approve once the items below are addressed, with an external crypto sign-off on the heuristics.** Focusing on net-new findings below; I won't repeat what's already covered in the prior reviews.

## Verified correct

- **DRY refactor is behavior-preserving.** `translated_subset_masks` is now the single source of truth shared by `recover_share` (signing-critical — affects *all* signing) and the Round-5 guard. Diffed against the old inline logic: ascending active indices == old `sorted_active_indices`, identical perm fill, identical `pattern→perm[pos]` translation, same `current_i >= patterns.len()` guard. No regression.
- **Norm-bound math is consistent across Rust and Python.** `partial_secret_norm_bound` == `κ·beta_bound` in `compute_hyperball_params.py`: `var_eta = η(η+1)/3` is algebraically equal to `sigt² = ((2η+1)²−1)/12`, and `dim`, `ν`, `num_secrets = ⌈C(n,t−1)/t⌉` all match. `√τ` is on both sides of the guard, so the new CI regression assertion (`overshoot ≤ κ`) checks exactly the quantity κ is selected against.
- **Zero-sum noise is exact.** In `add_mean_subtracted_noise`, `Σ sub = m·base + remainder = total` (`div_euclid`/`rem_euclid`), so `Σ N_j = 0` — the integer secret identity is preserved. `sample_split_noise_coeff` consumes exactly one byte; `split_noise_threshold` clamps to `[1,127]` so `2T ≤ 254 < 256` (no band overflow). Determinism across old-subset peers holds (PRF-keyed; deterministic byte consumption).

## Net-new findings

**1. `COSET_RESHARING_SPEC.md` is referenced 11× but does not exist in the repo.** It's cited in `protocol.rs` (×3), `README.md` (×3), `resharing_tests.rs` (×2), `SECURITY_PROOF.md`, `config.rs`, and `compute_hyperball_params.py` — including as the authoritative source for the v5 "Option A" design and the entire roadmap to remove the 4-of-6 `K=1600` tax ("Option B/C targets κ=1/K=350"). `git ls-files | rg -i coset` returns nothing. Either commit the file or replace the references with inline rationale; right now the most load-bearing design claim points at a missing file.

**2. 4-of-6 — the config this PR exists to enable — has zero default-CI coverage of its acceptance/κ path.** `test_reshares_and_sign_4_of_6`, `test_recovered_partial_variance_4_of_6`, `test_coefficient_distribution_4_of_6`, and `test_dealer_sign_4_of_6` are all `#[ignore]`d. So the new κ-margin regression assertion runs for `(2,2)/(2,3)/(2,4)/(3,5)` but **not** `(4,6)` — which carries the thinnest margin (measured 1.163 → κ=1.25, ~7.5%). Suggest a nightly/manual job for it, or an explicit note that its margin is unverified in CI.

**3. κ margins rest on un-CI'd measurement.** κ comes from Monte-Carlo in `compute_hyperball_params.py` (not run in CI) plus Rust `test_recovered_partial_variance_*` (4-of-6's is `#[ignore]`d), over limited seeds/reshares. If the true worst-case overshoot exceeds κ this is a **liveness** failure (honest reshares rejected), not a security break — but it argues for wiring at least the python sanity check (and `keyhiding_conditional.py`) into CI, or pinning its output.

**4. Scope creep in `hdwallet/src/wormhole.rs`.** Comment-only, but it (a) flips a contradictory prior comment ("4-bytes/felt, 8 felts" → "8-bytes/felt, 4 felts") and (b) asserts the non-injective `bytes_to_digest` encoding is "harmless … a collision would only alias the holder's own secret." That's a real security argument about address derivation, unrelated to resharing, and deserves its own review rather than riding in here. The DeepWiki badge is likewise unrelated (harmless).

## Carried-over (still present)

- **Round-3 peer-abort liveness asymmetry.** `verify_peer_dealer_commitments()?` returns a local `Err` out of `poke()` without setting `Failed` or broadcasting (unlike Round-5). Halting before sending one's own Round-4 shares is the correct security call, but new-only members hang until a transport timeout — confirm the `near-mpc` runner turns this `Err` into a global abort/timeout.
- **`√τ` is the average-case (Gaussian-heuristic) amplification, not worst-case (`τ`).** The guard faithfully enforces the proof's *own* condition (B and the radii are defined with `√τ`), but it is not a worst-case bound on `‖c·u‖₂`. This — plus the ≤12-bit hint-MLWE hiding loss of a-posteriori sharing and the κ radius enlargement — is what genuinely needs an external cryptographer's sign-off, given this is committee-handoff code for signing keys.
- **Minor DRY remnant:** dealer-commitment lookup is still duplicated between `verify_peer_dealer_commitments` and `verify_and_aggregate_new_shares`; `compute_sharing_patterns` is recomputed per signing-set inside the guard loop (micro).

n13:
Minor changes requested, see above

n13 (commented):
## Review: Resharing hardening — re-review (head `4260d4b`)

Re-reviewed the two commits since my last pass (`9836645` *unignore tests, fix stale comments*, `4260d4b` *dedupe more code, mark state as failed*). **This iteration closes every actionable item I raised. Engineering is merge-ready; the remaining gate is an external cryptographer's sign-off on the heuristics, which is not a code issue.** CI is green on the head commit (all 9 checks, incl. test workspace on ubuntu + macos).

### Resolved this round
- **Missing `COSET_RESHARING_SPEC.md` (the load-bearing finding)** — all 11 references removed and replaced with inline rationale across `protocol.rs`, `config.rs`, `README.md`, `SECURITY_PROOF.md`, `compute_hyperball_params.py`, and the tests. `git grep COSET_RESHARING_SPEC` → zero hits.
- **(4,6) had zero default-CI coverage** — un-ignored `test_reshares_and_sign_4_of_6`, `test_recovered_partial_variance_4_of_6`, `test_coefficient_distribution_4_of_6`, `test_dealer_sign_4_of_6`, and `test_reshares_and_sign_3_of_5`. The thinnest-margin config now exercises its acceptance/κ path in CI.
- **κ regression assertion** — `analyze_recovered_partials` now asserts `guard_overshoot <= resharing_norm_enlargement(threshold, parties)`, reading κ from the *same* source the guard uses (now `pub` + re-exported). This regression-protects exactly the quantity κ is selected against, including (4,6).
- **DRY: dealer-commitment lookup** — extracted `dealer_round3_for`, shared by `verify_peer_dealer_commitments` (Round 3) and `verify_and_aggregate_new_shares` (Round 5). Behavior-preserving (`None` and `dealer == me` paths resolve identically).
- **Micro: `compute_sharing_patterns` per-signing-set recompute** — hoisted out of the guard loop.
- **Round-3 peer-abort liveness** — abort now sets `ResharingState::Failed(...)` before returning `Err`, so the halt is terminal and observable via `is_failed()` (matching the Round-5 abort). The silent no-broadcast behavior remains a deliberate security choice, now documented with the near-mpc timeout path.

I re-verified the signing-critical `translated_subset_masks` extraction against the old inline `recover_share` logic (ascending active indices, identical perm fill, identical pattern→`perm[pos]` translation, same `current_i >= patterns.len()` guard); the OR-fold `signing_mask` is order-independent so dropping the explicit sort is safe. No regression, and the signing suite is green.

### Net-new nit (minor, from `4260d4b`)
Inserting `dealer_round3_for` displaced a doc comment: the *"Old-subset peer verification for Round 3 dealer commitments"* block now sits above the lookup helper, and `verify_peer_dealer_commitments` is left undocumented. Not a functional bug (clippy/doc CI passes) — just move that prose down one function. One-line fix, non-blocking.

### Remaining carry-overs (not blocking)
- **External crypto sign-off** is the real gate: √τ is the Gaussian-heuristic (not worst-case τ) amplification, plus the ≤12-bit hint-MLWE hiding loss and the κ radius enlargement. The code faithfully enforces the proof's *own* condition; the proof itself needs a cryptographer.
- **κ rests on single-seed Monte-Carlo** — the Rust assertion now guards the enforced quantity, but `keyhiding_conditional.py` / `compute_hyperball_params.py` still aren't wired into CI and the (4,6) variance test uses one seed.
- **Heavier default `cargo test`** — the un-ignores add several K=1600 (4,6) runs to the default suite (~9 min in CI). Deliberate coverage tradeoff; local runs are now slower.
- **`hdwallet/src/wormhole.rs`** comment-only change still rides along (address-derivation argument, unrelated to resharing). Harmless, arguably its own PR.

n13 (approved):
GTG

PR #90, opened 2026-06-18
dilithium: stream matrix A to cut ML-DSA-87 peak stack for embedded

n13:
## Summary
- ML-DSA-87 keygen/sign/verify currently materialize the full ExpandA matrix (`[Polyvecl; K]`, ~56 KB) on the stack. On the Keystone3 / ForgeBox (Cortex-M, `thumbv7em-none-eabihf`) target this overflows the task stacks and **freezes the device** when deriving a Quantus address or signing.
- Add `matrix_pointwise_montgomery_streamed`, which computes `t = A * v` while regenerating each `A[i][j]` from `rho` on the fly. Peak extra working set drops from ~56 KB to ~2 KB (two polynomials). Accumulation order is identical to the materialized path, so output is **bit-for-bit unchanged**.
- `SigningContext` now stores the 32-byte public seed `rho` instead of the expanded matrix; `A` is streamed per rejection-sampling attempt (small recompute, large stack saving).

## Measured impact (peak stack)
| op | before | after |
|---|---|---|
| `sign` (thumbv7em frame) | — | ~109 KB |
| `verify` | — | ~50 KB |
| `keypair` | — | ~39 KB |
| `sign` (host painted-stack probe) | ~309 KB | ~157 KB |
| `keygen` (host) | ~173 KB | ~69 KB |
| `verify` (host) | ~188 KB | ~84 KB |

## Pre-flight stack tooling (new)
- `dilithium/examples/stack_probe.rs` — painted-stack high-water probe (`psm` dev-dependency) for host peak-stack numbers.
- `stack-check.sh` — builds dilithium for `thumbv7em-none-eabihf` with `-Z emit-stack-sizes` and reports per-function frames via `llvm-readobj --stack-sizes`, with an optional KB budget check to catch regressions (e.g. accidentally re-materializing the matrix) before flashing.

## Correctness
- Verified **byte-identical** keygen/sign/verify against the NIST KAT (`tests/` integration suite).

## Test plan
- [ ] `cargo test` (workspace, incl. KAT integration tests) passes
- [ ] `cargo run --release --example stack_probe -p qp-rusty-crystals-dilithium`
- [ ] `./stack-check.sh` reports expected frames
- [ ] On-device: select Quantus, enter PIN, derive address + sign a tx without freeze

## Notes
HD derivation / address generation continue to use rusty-crystals unchanged. A deeper BRS22 buffer-reuse refactor (to fit small SRAM task stacks directly, without the dedicated PSRAM crypto task on the firmware side) is a possible follow-up.

---

> [!NOTE]
> **Medium Risk**
> Core cryptographic paths changed (matrix multiply implementation) but claim bit-identical results; signing does more recomputation per attempt. Medium risk due to embedded-critical stack fix and algorithm path swap, mitigated by KAT/tests and equivalence intent.
> 
> **Overview**
> **ML-DSA-87 keygen, sign, and verify** no longer allocate the full ExpandA matrix (~56 KB on stack). A new **`matrix_pointwise_montgomery_streamed`** computes `A·v` by expanding each `A[i][j]` from `rho` on the fly (~2 KB scratch), with the same accumulation order as before so outputs stay **bit-identical**.
> 
> **Signing** keeps the 32-byte public seed `rho` in `SigningContext` instead of a materialized matrix; each rejection attempt re-streams `A` from `rho`.
> 
> **Tooling:** `stack_probe` (host painted-stack peak measurement via `psm`) and **`stack-check.sh`** (Cortex-M `thumbv7em` build with `-Z emit-stack-sizes`, optional KB budget) help catch stack regressions before flashing embedded targets.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 24259fa1929fef80a27783c339b4ed286f1f94f8. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


illuzen:
v12 plz audit

n13 (commented):
## Audit Review: Constant-Time Properties & Correctness

### Verdict: **Approve with minor notes**

The streaming approach is sound — it does **not** weaken constant-time properties, and the arithmetic equivalence claim holds. The change is appropriate for the embedded stack constraint it targets.

---

### 1. Arithmetic Equivalence — ✅ Verified

The streamed function faithfully replicates the original two-step path:

**Original:**
```
matrix_expand:                 mat[i].vec[j] = uniform(rho, (i<<8)+j)
l_pointwise_acc_montgomery:    t[i] = Σ_j mat[i][j] · v[j]
```

**Streamed:**
```
for each row i:
  a_ij = uniform(rho, (i<<8)+0);  t_i  = a_ij · v[0]
  for j in 1..L:
    a_ij = uniform(rho, (i<<8)+j); t_i += a_ij · v[j]
```

- Nonce encoding `(i << 8) + j` matches exactly. Max value = `(7 << 8) + 6 = 1798`, well within `u16`.
- Accumulation order (first element directly, then add remaining) matches `l_pointwise_acc_montgomery`.
- Same SHAKE128 stream → same A coefficients → **bit-for-bit identical output**. KAT verification confirms this.

### 2. Constant-Time Analysis — ✅ No Regression

The critical question: does re-expanding A from `rho` on every signing attempt leak secret-dependent timing?

**No, because the data flow is clean:**

| Operation | Touches secrets? | Constant-time? |
|---|---|---|
| `poly::uniform(a_ij, rho, nonce)` | No — `rho` is public (in the public key) | Variable (rejection sampling), but **irrelevant** — depends only on public `rho`, not on secrets |
| `poly::pointwise_montgomery(t_i, a_ij, v[j])` | `v` may contain secret-derived NTT values (e.g. `s1hat` in keygen, `y_hat` in signing) | **Yes** — fixed N iterations, Montgomery reduction is data-independent |
| `poly::add_ip(t_i, prod)` | Yes | **Yes** — fixed N iterations, simple addition |
| NTT / invNTT | Yes | **Yes** — fixed butterfly structure, Montgomery-based |

**Key insight:** `poly::uniform` (SHAKE128 rejection sampling via `rej_uniform`) has data-dependent timing, but this is a **pre-existing** property shared with the original `matrix_expand` path. Critically, it depends only on the public seed `rho` — for a given `rho`, the timing is deterministic across all signing attempts, so it cannot distinguish which attempt succeeded.

The operations that process secret data (`pointwise_montgomery`, `add_ip`, NTT) are all constant-time with fixed loop bounds.

The signing loop's `MIN_SIGNING_ATTEMPTS = 16` batching (which masks rejection count) is preserved unchanged.

**Per-path summary:**
- **Keygen:** A·s1 — `s1hat` is secret but only enters through constant-time `pointwise_montgomery`. A expansion depends only on public `rho`. ✅
- **Sign:** A·ŷ per attempt — `ŷ` (NTT of masking vector) enters through constant-time ops. A expansion is deterministic for fixed `rho`. ✅
- **Verify:** A·z — both `rho` and `z` are public (from pk and signature). No secrets involved at all. ✅

### 3. Issues Found

#### 3a. Performance cost in signing (informational, not a bug)

Each signing attempt now re-expands A: `K × L = 56` SHAKE128 polynomial generations per attempt. With `MIN_SIGNING_ATTEMPTS = 16`, that's **896** SHAKE128 expansions per `sign()` call minimum, versus 56 (once) before. This is the intended tradeoff for ~56 KB less peak stack, but it's a meaningful host-side performance regression (~16× more A expansion work).

On the embedded target this is the right call. If host performance matters, a feature gate (`#[cfg(feature = "low-stack")]`) could let callers opt in, keeping the materialized path as default for unconstrained targets.

#### 3b. `matrix_expand` + `matrix_pointwise_montgomery` now dead in production (minor)

The original functions remain in `polyvec.rs` but are no longer called from any production path (`keypair`, `signature`, `verify`). They're still exercised by unit tests, which is fine — they serve as reference implementations. No action needed, but worth being aware if pruning dead code later.

#### 3c. CodeQL alerts are false positives

The two "hard-coded cryptographic value" alerts on `(i << 8) as u16` and `((i << 8) + j) as u16` are false positives. These are deterministic index-derived nonces as specified by the ML-DSA/FIPS 204 ExpandA definition — they are not secret nonces.

### 4. Conclusion

The PR achieves its goal cleanly: ~56 KB peak stack reduction with no change to cryptographic output and no weakening of constant-time properties. The arithmetic path is equivalent, secrets never flow through variable-time operations, and the timing countermeasures in the signing loop are preserved. The only cost is re-computation of public matrix A per attempt — an acceptable tradeoff for the embedded target.

illuzen (approved):

PR #91, opened 2026-06-19
Add acvp vectors and tests for them

illuzen:
> [!NOTE]
> **Low Risk**
> Changes are confined to tests, dev-dependencies, and documentation; signing/verification library code paths are exercised but not modified.
> 
> **Overview**
> Adds a **test-only ACVP known-answer harness** so ML-DSA-87 can be checked bit-for-bit against vendored NIST vectors (keyGen, sigGen, sigVer) for the FIPS 204 **internal** `Sign_internal` / `Verify_internal` path via `crate::sign::{keypair, signature, verify}`.
> 
> The harness lives in `dilithium/src/acvp.rs`, loads JSON from `dilithium/tests/acvp_vectors/`, and is wired with `#[cfg(test)]` plus `extern crate std` only for tests so production stays `#![no_std]`. **Dev-dependencies** `serde_json` and `hex` support parsing vectors.
> 
> **`NIST_VALIDATION.md`** documents CAVP vs CMVP vs MPTC, how to run `cargo test -p qp-rusty-crystals-dilithium --lib acvp`, vector refresh with `jq`, and optional follow-ups (external-interface vectors, CI).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9957d6539163f82b2139807bfa979e9f1781cb1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13:
I reviewed PR #91 thoroughly, including checking out the branch and actually running the harness. Here's my assessment.

## Verdict: Approve (solid, well-scoped). Only minor, optional nits.

This adds a `#[cfg(test)]`-only ACVP known-answer harness validating ML-DSA-87 bit-for-bit against vendored NIST vectors. No production library logic changes — only test code, dev-deps, `Cargo.lock`, and docs.

## What I verified

- **All vectors pass, locally.** `cargo test -p qp-rusty-crystals-dilithium --lib acvp`: keyGen 25 / sigGen 30 / sigVer 15. Full lib suite: 90/90, no regressions. `cargo fmt --check` and `cargo clippy --tests` are both clean.
- **The key correctness question: does the custom rejection-sampling loop break byte-exact reproduction?** No. The signer batches `MIN_SIGNING_ATTEMPTS = 16` for timing reasons, but the `!signature_found` guard keeps the *lowest* valid nonce and the outer loop continues with a monotonic nonce, so output matches FIPS 204's "first valid κ" semantics:

```435:440:dilithium/src/sign.rs
			if all_conditions_met && !signature_found {
				valid_challenge.copy_from_slice(&dummy_output[..params::C_DASH_BYTES]);
				valid_signature_z = signature_z;
				valid_hint_h = hint_vector_h.clone();
				signature_found = true;
			}
```

That this matters is confirmed empirically — all 30 sigGen vectors (deterministic + hedged) reproduce exactly.
- **Vector provenance claim ("unmodified except filtering") holds.** The vendored files' top-level shape is exactly `{algorithm, mode, revision, testGroups}` — what the documented `jq` recipe produces — and groups are correctly filtered to `ML-DSA-87`, `signatureInterface == "internal"`, `externalMu == false`. Per-test fields are untouched.
- **`no_std` is preserved.** `extern crate std` and `mod acvp` are both `#[cfg(test)]`; production builds stay `#![no_std]`.
- Mapping is correct: deterministic groups → `hedge = None` (→ `rnd = 0^32`), hedged → vector `rnd`; sigVer's length-mutated negative cases correctly resolve to `false`.

## Minor / optional findings (none blocking)

1. **Dev-dep pin inconsistency.** Every existing dev-dep is exact-pinned (`=0.5.1`, `=0.10.1`, `=0.7`, `=0.1.31`) but the new ones aren't:

```30:32:dilithium/Cargo.toml
# NIST ACVP known-answer-test harness (tests only; does not affect the no_std lib).
serde_json = "1"
hex = "0.4"
```

`Cargo.lock` is committed so builds are reproducible regardless, but for consistency consider `=1.0.x` / `=0.4.x`.

2. **DRY (per your repo rule).** The three test fns repeat the same skeleton (load → iterate groups/tests → `count` → `assert!(count > 0)` → `println!`). A small helper taking a per-test closure would remove it. Borderline — readability is fine as-is — so optional.

3. **No integrity pin on the upstream source.** Regeneration pulls from `master` of `usnistgov/ACVP-Server` with no commit SHA / checksum. A pinned upstream ref or a SHA256 of the source files would make the provenance reproducible. Optional.

4. **Placement nit:** vectors live under `tests/acvp_vectors/` but are consumed by a `--lib` unit test (`src/acvp.rs`), not an integration test in `tests/`. Works fine via `CARGO_MANIFEST_DIR`; just slightly unconventional.

## Strengths

- Genuinely valuable: byte-exact NIST conformance for keyGen/sigGen (incl. the rejection loop) plus accept/reject sigVer cases.
- Clean `#[cfg(test)]` scoping keeps production `no_std`.
- Fail-loud `panic!`/`expect` everywhere — aligns with your "fail early, no silent failures" rule.
- `NIST_VALIDATION.md` is accurate and genuinely useful (CAVP vs CMVP vs MPTC, regeneration recipe, honest scope boundaries — internal-only, ML-DSA-87 only, KAT not a proof).

One forward-looking note matching the PR's own "next steps": the harness covers only the internal `Sign_internal`/`Verify_internal` path, so the public `ml_dsa_87` API's domain-separation prefix/context handling is still untested — worth a follow-up with the external-interface vectors.

n13 (approved):
GTG!

PR #93, opened 2026-07-07
Better constant time

illuzen:
## Summary

This PR reworks the constant-time posture of the `dilithium` crate based on a review of
what actually needs protection under the FIPS 204 security analysis. It removes expensive
countermeasures that only disguised **public** information, closes the remaining
secret-dependent timing gaps, and replaces the lost dudect harness with one that measures
the right thing.

Net effect: **signing is ~22x faster** (2.82 ms → 128 µs on Apple Silicon, criterion
`Dilithium signing` bench), signatures are **byte-identical** (all ACVP KATs pass
unchanged), and every secret-consuming component now passes dudect with |max t| well
under the 5.0 leakage threshold.

## Motivation

The previous implementation tried to make *end-to-end* signing time look constant to a
dudect test that compared fixed vs. random keys. But ML-DSA is Fiat–Shamir with aborts:
the number of rejection-sampling attempts is independent of the long-term secret key and
is treated as public in the FIPS 204 analysis (reference C, PQClean, and AWS-LC all use
plain early-exit loops). Chasing that non-leak led to three expensive countermeasures —
while several places that touch actual secrets still had data-dependent branches, and one
fix had introduced a secret-dependent division (the KyberSlash bug class). Meanwhile the
documented dudect harness (`--features ct-testing --bin ct_bench`) no longer existed, so
none of the claims were reproducible.

### Commit 1 — drop countermeasures that only masked the public abort count

- `sign.rs`: replace the 16-attempt batched rejection loop (`MIN_SIGNING_ATTEMPTS`) with
  a standard early-exit loop. Removes the per-attempt dummy `pack_sig` call, the dummy
  output buffer, and the data-dependent `Polyveck::default()` in the `safe_hint` branch.
  All four rejection checks are still evaluated unconditionally, so a rejected attempt
  reveals only *that* it was rejected, not which bound failed. The nonce-overflow assert
  is kept.
- `poly.rs::challenge()`: restore plain rejection sampling (was: one dummy Keccak
  permutation per candidate byte, ≥480 permutations per challenge vs. ~2). Its input
  `c~ = H(μ, w1)` is a hash output — published in the signature for accepted attempts,
  and unexploitable (preimage of an unpublished hash) for rejected ones.

### Commit 2 — close the real secret-dependent gaps

- `poly.rs::rej_eta`: `ctr % alen` → `ctr.min(alen - 1)`. The modulo was a hardware
  division with a secret-derived operand (KyberSlash class); `min` compiles to a
  conditional move. Semantics identical: `ctr` can only reach `alen` when the store mask
  is already zero, making the clamped access a self-assignment either way.
- `poly.rs::check_norm`: data-dependent branch → `result |= t >= b` (the code now matches
  its own comment).
- `rounding.rs::make_hint`: if/else on secret-derived `w0 − c·s2 + c·t0` values →
  sign-bit arithmetic. Hints of *rejected* attempts are never published, so this path is
  secret. Verified branchless/vectorized in the emitted assembly.
- `packing.rs::pack_sig`: the branchless masked store was undermined by
  `if is_nonzero { k += 1 }` and an if/else write-index select on the same hint data →
  arithmetic increment and `min` clamp.

### Commit 3 — new dudect harness + docs

- New `dilithium/examples/ct_bench.rs` (same target layout as qp-poseidon). Instead of
  timing whole `sign()`/`keygen()` calls — which mostly measures the public abort count
  and ExpandA's rejection sampling on the public ρ — it isolates each secret-consuming
  component, fixed secret vs. random secret:

  | Test | Component | max t (pass < 5) |
  |---|---|---|
  | `keygen_s1s2_sampling` | s1/s2 eta sampling from ρ′ | ~2.1 |
  | `rej_eta_one_block` | raw eta sampler, one SHAKE block | ~2.7 |
  | `sign_sk_expansion` | sk unpack + NTT of s1/s2/t0 | ~1.8 |
  | `sign_mask_expansion` | ExpandMask for y | ~1.5 |
  | `sign_norm_check` | norm check on z = y + c·s1 | ~1.6 |
  | `sign_make_hint` | hint computation | ~1.7 |
  | `ntt_pointwise` | NTT + pointwise multiply | ~2.6 |

  Harness design: both classes do identical per-sample prep (RNG fill + memcpy into one
  shared working buffer) so RNG cost and cache state cannot differ between classes;
  sub-microsecond ops are batched above the ~41 ns timer tick; no 8 MB cache flush per
  sample, so sample counts are 10–40x higher than the old harness for the same wall time.

- `CONSTANT_TIME_TESTING.md` rewritten around "what to measure and why": documents the
  abort-count argument, what is deliberately excluded (`challenge`, whole-call timing,
  `pack_sig` — only called for the accepted attempt with published data, `verify`), and
  replaces the stale run instructions with the working command:

  ```bash
  cargo run --release -p qp-rusty-crystals-dilithium --example ct_bench
  ```

## Test plan

- [x] `cargo test -p qp-rusty-crystals-dilithium` — 90 unit tests + ACVP keygen/siggen/sigver KATs pass (signatures byte-identical, so the loop restructure and branchless rewrites are behavior-preserving)
- [x] `cargo test -p rusty-crystals-integration-tests` — NIST KAT integration tests pass
- [x] `cargo run --release -p qp-rusty-crystals-dilithium --example ct_bench` — all components |max t| < 5 across repeated runs
- [x] `cargo clippy --release -p qp-rusty-crystals-dilithium --all-targets` — clean
- [x] Emitted assembly inspected for `rej_eta` and `make_hint` (branchless: `csel`/`cset`/SIMD compares only)
- [x] Each commit compiles independently (checked for bisectability)
- [ ] Optional follow-up: run `ct_bench --continuous` on a quiet pinned-frequency machine for higher-confidence numbers; consider a scheduled CI job on a dedicated runner

## Notes for reviewers

- The known residual, inherent to the construction and shared with the reference
  implementation: `rej_eta`'s store address follows the acceptance counter (cache-line
  granularity, within a 4 KB buffer). dudect shows no signal; documented in the md.
- qp-poseidon's `CONSTANT_TIME_TESTING.md` has the same stale `--features ct-testing
  --bin ct_bench` instruction this PR fixes here; worth a small follow-up PR there.

---

> [!NOTE]
> **High Risk**
> Touches signing control flow and secret-handling primitives in a post-quantum signature crate; behavior is intended to stay KAT-identical but timing and side-channel posture change materially.
> 
> **Overview**
> Reworks ML-DSA-87 **constant-time** behavior: drops costly countermeasures that only hid the **public** Fiat–Shamir abort count, fixes remaining **secret-dependent** timing paths, and adds a **dudect** harness that measures isolated components instead of whole `sign()` / keygen.
> 
> **Signing (`sign.rs`)** uses a normal early-exit rejection loop again (no 16-attempt batching, dummy `pack_sig`, or deferred packing). All four rejection checks still run every attempt; acceptance uses bitwise `&` so a failure does not reveal which bound broke. Challenge bytes are written directly into the output buffer on accept.
> 
> **Crypto primitives:** `poly::challenge()` returns to plain variable-time rejection sampling (input is public hash material). `rej_eta` replaces `ctr % alen` with `ctr.min(alen - 1)` to avoid KyberSlash-style division timing. `check_norm` accumulates with `|=` instead of branching. `make_hint` and `pack_sig` hint packing are fully branchless (`min` clamp, arithmetic `k` increment).
> 
> **Tooling:** New `examples/ct_bench.rs` plus `CONSTANT_TIME_TESTING.md` updates document what to time (per-component fixed vs random secrets) and what to exclude (whole-call timing, `challenge`, etc.).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d23326f6385141516208d87db576d3b646a77f74. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dilithium/CONSTANT_TIME_TESTING.md
+++ b/dilithium/CONSTANT_TIME_TESTING.md
@@ -1,128 +1,130 @@
 # Constant-Time Testing for qp-rusty-crystals-dilithium
 
-This project uses [dudect-bencher](https://github.com/rozbb/dudect-bencher) to test whether the Dilithium ML-DSA-87 digital signature and key generation operations execute in constant time, resisting timing side-channel attacks.
+This project uses [dudect-bencher](https://github.com/rozbb/dudect-bencher) to test that the
+secret-dependent components of the ML-DSA-87 implementation execute in constant time.
 
 ## Quick Start
 
-Run all constant-time tests:
+```bash
+cargo run --release -p qp-rusty-crystals-dilithium --example ct_bench
+```
+
+Run a single test (with continuous sampling until interrupted):
 
 ```bash
-cd dilithium
-cargo run --release --features ct-testing --bin ct_bench
+cargo run --release -p qp-rusty-crystals-dilithium --example ct_bench -- --continuous sign_norm_check
 ```
+
+## What we measure — and what we deliberately do not
+
+This is the important part, and it is where a naive test setup goes wrong.
+
+ML-DSA signing is **Fiat-Shamir with aborts**: the signer retries (on average ~4 times for
+ML-DSA-87) until a candidate signature passes four rejection checks. The *number of attempts*
+is independent of the long-term secret key and is treated as **public information** in the
+FIPS 204 security analysis — every mainstream implementation (reference C, PQClean, AWS-LC)
+uses a plain early-exit rejection loop. End-to-end signing time therefore varies from call to
+call *by design*.
+
+Consequence: timing the whole `sign()` call with dudect (fixed key vs. random key, or fixed
+message vs. random message) produces a large t-statistic that looks like a leak but is not —
+dudect is simply detecting the public abort count. The same applies to whole
+`Keypair::generate()` calls: expanding the matrix A uses rejection sampling on `rho`, and
+`rho` is published in the public key.
+
+What must NOT depend on secrets is the work done **inside** each attempt and around it.
+The harness (`examples/ct_bench.rs`) therefore isolates each secret-consuming component and
+compares a fixed secret input (Class Left) against fresh random secret inputs (Class Right):
+
+| Test                   | Component under test                                        |
+|------------------------|-------------------------------------------------------------|
+| `keygen_s1s2_sampling` | Sampling secret vectors s1/s2 from rho' (keygen)            |
+| `rej_eta_one_block`    | Raw eta rejection sampler over one SHAKE block              |
+| `sign_sk_expansion`    | Secret-key unpack + NTT of s1/s2/t0 (per-sign setup)        |
+| `sign_mask_expansion`  | Mask vector y expansion from rho' (ExpandMask)              |
+| `sign_norm_check`      | Infinity-norm rejection check on z = y + c·s1               |
+| `sign_make_hint`       | Hint computation from secret-derived w0/w1                  |
+| `ntt_pointwise`        | NTT + pointwise multiplication on secret operands           |
+
+Deliberately excluded:
+
+- **Whole `sign()` / keygen calls** — timing varies with public information (abort count,
+  `rho`), see above.
+- **`poly::challenge()`** — variable-time by design. Its input `c~ = H(mu, w1)` is a hash
+  output: published in the signature for accepted attempts, and unexploitable (a preimage
+  of an unpublished hash) for rejected ones. No secret-key material flows into it.
+- **`packing::pack_sig()`** — only called for the accepted attempt, so its inputs are
+  exactly the published signature bytes. (The hint loop is branchless anyway as
+  defense-in-depth, but its hint-weight-dependent store footprint is not a secret channel.)
+- **`verify()`** — operates exclusively on public data.
 
 ## Understanding Results
 
-Each test shows output like:
+Each test prints a line like:
+
 ```
-test_keypair_generation_ct        : n == +0.005M, max t = +1.83452, max tau = +0.08201, (5/tau)^2 = 3715
+bench sign_norm_check ... : n == +0.149M, max t = +1.53271, max tau = +0.00397, (5/tau)^2 = 1587037
 ```
 
-- **n**: Number of measurements taken (in millions)
-- **max t**: Raw t-statistic from Welch's t-test
-- **max tau**: Normalized t-statistic (more reliable for large sample sizes)
-- **(5/tau)²**: Detection threshold - higher numbers are BETTER
+- **n**: number of measurements used (in millions, after outlier trimming)
+- **max t**: worst Welch t-statistic over all measurement crops
+- **max tau**: t normalized by sqrt(n) — comparable across sample sizes
+- **(5/tau)²**: roughly how many measurements would be needed to confirm a leak
 
-### Interpretation Guidelines
+Interpretation:
 
-- **|max t| < 5.0**: ✅ GOOD - No timing leakage detected
-- **|max tau| < 0.1**: ✅ GOOD - No timing leakage detected
-- **|max tau| ≥ 0.1**: ⚠️ CONCERN - Potential timing side-channel detected
-- **Higher (5/tau)² values**: ✅ GOOD - Higher numbers indicates more measurements would be needed to detect any leakage
-
-## Tests Included
-
-### Key Generation
-- `test_keypair_generation_ct` - Tests ML-DSA-87 keypair generation with different entropy sources
-
-### Signing Operations  
-- `test_signing_small_ct` - Small messages (32 bytes)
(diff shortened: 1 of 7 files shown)


n13 (approved):
## Review: Approve

This is a well-reasoned, high-quality rework of the constant-time posture. The central argument — that the Fiat-Shamir abort count is public under the FIPS 204 security model and the expensive countermeasures were hiding the wrong thing — is sound and well-supported by how the reference C, PQClean, and AWS-LC implementations handle the same construction.

### What this gets right

**Fixing real secret-dependent leaks (Commit 2):** This is the most valuable part of the PR. Four genuine timing gaps are closed:

- **`rej_eta`: `ctr % alen` → `ctr.min(alen - 1)`** — The modulo was a textbook KyberSlash-class bug. Division latency can be operand-dependent on ARM (and some x86), and `ctr` is secret-derived (acceptance counter from secret polynomial coefficients). `min` compiles to `csel`/`cmov`. The semantics argument is correct: when `ctr == alen`, the store mask is already zero, so the clamped write is a self-assignment.

- **`check_norm`: branch → `result |= t >= b`** — The old code literally had a comment saying "bitwise OR" while using a branch. The new code matches the stated intent and avoids a branch on the secret-derived absolute value of z = y + c·s1 coefficients.

- **`make_hint`: if/else → sign-bit arithmetic** — Verified the arithmetic manually:
  - `gt = (GAMMA2 - a0) >> 31` → −1 iff `a0 > GAMMA2` ✓
  - `lt = (a0 + GAMMA2) >> 31` → −1 iff `a0 < −GAMMA2` ✓
  - `eq = !((t | t.wrapping_neg()) >> 31)` → −1 iff `t == 0` (i.e. `a0 == −GAMMA2`) ✓ — `wrapping_neg` is the right call to avoid debug-mode panic on `i32::MIN`
  - `a1_nonzero` same pattern ✓
  - Final OR matches the original three-way condition exactly ✓

- **`pack_sig`: branchless counter** — `k += is_nonzero as usize` and `k.min(OMEGA - 1)` are clean replacements for the undermined "branchless" store that still had `if is_nonzero { k += 1 }`.

**Simplification of `store_mask` in `rej_eta`:** The old `let store_mask = -((inc_ctr & has_space) as i32)` had a redundant `& has_space` since `inc_ctr` already incorporates `has_space` via `nibble_valid & has_space`. Simplifying to `-(inc_ctr as i32)` is correct and cleaner.

**Dropping the abort-count countermeasures (Commit 1):** The `MIN_SIGNING_ATTEMPTS = 16` batching, dummy `pack_sig`, and the 480+ dummy Keccak permutations in `challenge()` were all protecting information that is public by construction. The 22x speedup is a direct consequence of removing work that shouldn't have been there. The bitwise `&` joining the four conditions in the acceptance check preserves the non-short-circuit property, so a rejected attempt still doesn't reveal *which* bound failed.

**Writing challenge directly to `signature_output`:** Each iteration overwrites `signature_output[..C_DASH_BYTES]` via `generate_challenge_polynomial`, and `pack_sig(signature_output, None, ...)` correctly skips the challenge region since it's already in place from the accepted attempt. Clean.

**Dudect harness (Commit 3):** Well-designed: both classes do identical per-sample prep (RNG fill + memcpy into a shared working buffer), sub-µs ops are batched above the timer tick, no cache flush, and the tests target the right isolation boundaries. The 1000-slot output buffer for `rej_eta` testing with the explanation of why 256 slots would produce a false signal is a good catch.

### Minor observations (non-blocking)

1. **Compiler preservation caveat:** The branchless constructs in `make_hint`, `check_norm`, and `pack_sig` rely on the compiler not re-introducing branches (the standard caveat for any Rust/C CT implementation). The PR notes that `rej_eta` and `make_hint` were verified in emitted assembly on Apple Silicon — worth re-checking after any future LLVM/rustc upgrade or if targeting other architectures (e.g. the `thumbv7em` CI target).

2. **`use_hint` is still branchy:** `rounding::use_hint` uses `if hint == 0 ... else if a0 > 0 ... else ...`. This is fine because `use_hint` is only called in `verify()`, which operates on public data. Mentioning for completeness.

3. **`rej_eta` store address residual:** The known residual (store address follows the acceptance counter at cache-line granularity) is documented in the md and shared with the reference implementation. dudect shows no signal. Acknowledged.

4. **CodeQL alerts on the harness are false positives:** The `0` nonce parameters in `l_uniform_eta` / `k_uniform_eta` calls and the `attempt_nonce` initialization are standard counter starting values, not hard-coded cryptographic keys.

### Verdict

The security argument is sound, the fixes are correct and behavior-preserving (KATs unchanged), the harness tests the right things, CI is fully green, and the commit structure is clean and bisectable. **Approve.**

PR #95, opened 2026-07-10
V12 issues again

illuzen:
## Summary

This PR addresses a batch of security-review findings across the three crates
(`dilithium`, `hdwallet`, `threshold`). Each finding was verified, reproduced
with a test where practical, fixed, and re-tested. The changes fall into a few
themes:

- **Memory hygiene / zeroization** — ensure secret material (seeds, shares,
  K_S, signing keys, NTT accumulators, wormhole secrets) is scrubbed and cannot
  be silently copied.
- **Untrusted-input hardening (DoS)** — bound allocations, enforce message-size
  and length limits, and turn panics/overflows on attacker-controlled input into
  clean errors.
- **Cryptographic binding / domain separation** — bind protocol state (session
  SSID, threshold policy, `active_set`) into the values it must authenticate.
- **API-surface reduction** — stop exposing low-level, panic-prone, or
  secret-bearing helpers on the public surface.
- **Correctness / canonicalization** — reject non-canonical encodings that
  would otherwise alias distinct inputs to the same key/address.

No functional behavior changes for honest callers, but several public APIs were
tightened (see **Breaking changes** below).

### `dilithium`

- **Reject all-zero-`t1` public keys** to prevent a no-secret forgery class.
- **`Keypair::from_bytes` now verifies the public key matches the secret key**,
  re-deriving `t1`/`t0` from the secret half and rejecting inconsistent blobs.
  The `t1`/`t0` derivation is shared between keygen and this check to avoid
  divergence.
- **Zeroize freshly-sampled secret polynomials** (`s1`, `s2`, `t0`) after they
  are packed into the secret key in `keypair()`.
- **Enforce exact-size buffers for low-level polynomial codecs** (`z_pack`,
  `z_unpack`, `k_pack_w1`, etc.) via array-typed parameters and
  `slice::as_chunks`/`first_chunk`; internal-only codecs are now `pub(crate)`.
  This removes a panic-on-short-slice footgun from the public API.
- **Make vector secret samplers `pub(crate)`** to prevent nonce-reuse misuse,
  with feature-gated wrappers for benchmarking.
- **Zero the hint vector in `unpack_sig`** to prevent stale-hint carryover.
- **Absorb message slices instead of concatenating** in `sign`/`verify`,
  eliminating a large-message heap-copy DoS.
- **Prevent a `keccak_squeeze` hang** on an out-of-range `KeccakState.pos`.

### `hdwallet`

- **`WormholePair.secret` is now a move-only `SensitiveBytes32`** instead of a
  public `[u8; 32]`. The array was `Copy`, so `pair.secret` could silently
  duplicate the secret and defeat `ZeroizeOnDrop`. Reads now go through
  `secret.as_bytes()`.
- **`ChildNumber::from_str` requires canonical decimal** — rejects leading `+`,
  redundant leading zeros, and non-ASCII digits so distinct path strings can no
  longer alias to the same derived key.
- **NFKD-normalize mnemonic and passphrase** before BIP39 seed derivation.

### `threshold`

Memory hygiene:
- **Enforce zeroization of DKG/resharing transcript signer keys** by requiring
  `TranscriptSigner: ZeroizeOnDrop`; dropping the config now provably scrubs the
  long-term signing key.
- **Keep queued DKG Round 1 privates as zeroizing structs** (not raw `Vec<u8>`),
  and **zeroize NTT accumulators and dealer share material**.
- **Redact the `SendPrivate` payload in `DkgAction`'s `Debug`** so K_S is not
  leaked to logs/telemetry.
- **Remove secret-extraction APIs from `PrivateKeyShare`'s public surface.**

Untrusted-input hardening:
- **Bound Borsh deserialization before allocation** for `Round2Broadcast`,
  `Round3Broadcast`, `Round4Broadcast`, and `ResharingCertificate`.
- **Enforce `MAX_MESSAGE_SIZE` on the direct signer path.**
- **Enforce exact reveal-set and atomic aggregation in `round3_respond`** so a
  replayed/duplicated Round 2 reveal cannot poison in-progress signing state.
- **Validate `run_local_dkg` input lengths** instead of `unwrap()`/panicking.
- **Reject out-of-range partial-PK coefficients** in `pack_combined_pk` to
  prevent an `i32` overflow (panic in debug / silent wrap in release) during
  public-key aggregation.
- **Harden and unpublish `generate_subsets_of_size`** (u16 overflow / input
  validation; no longer re-exported at the crate root).

Cryptographic binding:
- **Bind dealer keygen to the `(t, n)` threshold policy and subset mask**, so
  distinct policies with the same subset count no longer produce identical
  public keys.
- **Bind DKG Round 1 randomness to the session SSID**, so retries with a fresh
  session nonce don't replay randomness.
- **Bind `active_set` into resharing certificate acceptance signatures**
  (`ACCEPT_DOMAIN` bumped to v2), so it can no longer be forged in an otherwise
  valid certificate.
- **Reject non-canonical subset masks in the public-key-preservation check.**

Correctness:
- **Use non-overflowing magnitude (`unsigned_abs`) for the subshare bound
  check**, fixing an `i32::MIN` edge case.

## Breaking changes

- `dilithium`: several low-level `poly`/`polyvec`/`packing` helpers are now
  `pub(crate)` or take exact-size array references; `Keypair::from_bytes` now
  rejects inconsistent `sk || pk` blobs.
- `hdwallet`: `WormholePair.secret` is a `SensitiveBytes32` (read via
  `.as_bytes()`), not a `[u8; 32]`. Downstream consumers (e.g. `quantus-cli`,
  the Quantus SDK) need trivial `.as_bytes()` updates.
- `threshold`: `TranscriptSigner` now requires `ZeroizeOnDrop`;
  `generate_subsets_of_size` is no longer publicly re-exported; `pack_combined_pk`
  is now fallible; `PrivateKeyShare`'s secret-extraction methods were removed.

## Test plan

- [x] `cargo test -p qp-rusty-crystals-dilithium`
- [x] `cargo test -p qp-rusty-crystals-hdwallet`
- [x] `cargo test -p qp-rusty-crystals-threshold` (lib, integration, doctests)
- [x] `cargo clippy` clean on the touched crates
- [x] New regression tests added for each fix where a runtime test is meaningful
      (public-key/secret-key consistency, canonical child numbers, wormhole
      secret wrapper, transcript-signer zeroization, partial-PK overflow
      rejection, active_set binding, dealer policy binding, SSID randomness
      binding, bounded deserialization, oversized-message rejection, etc.).

---

> [!NOTE]
> **High Risk**
> Touches authentication-critical paths (key import, public-key validation, DKG/resharing binding, signing state) and changes public APIs; honest callers should behave the same but integrators must adapt to stricter types and new error cases.
> 
> **Overview**
> Security-review fixes across **dilithium**, **hdwallet**, and **threshold**: tighter secret handling, safer untrusted input, and stronger protocol binding.
> 
> **ML-DSA (`dilithium`)** rejects degenerate all-zero `t1` public keys and verifies that `Keypair::from_bytes` re-derives a matching public key from the secret half. Sign/verify hash `domain_prefix` and message as separate SHAKE absorbs (no full-message heap copy). `unpack_sig` clears hints before decode; Keccak squeeze no longer spins on bad `pos`; internal sign/verify APIs take an explicit domain prefix. Low-level codecs use exact-size buffers; secret vector samplers are crate-internal with a `ct-internals` feature for `ct_bench`.
> 
> **HD wallet** wraps wormhole secrets in move-only `SensitiveBytes32`, requires canonical hardened child indices in paths, and NFKD-normalizes mnemonic/passphrase before BIP39.
> 
> **Threshold** requires `TranscriptSigner: ZeroizeOnDrop`, redacts K_S in `DkgAction` debug output, and queues Round 1 privates as zeroizing structs. Borsh paths use bounded/incremental reads; direct signing enforces `MAX_MESSAGE_SIZE`; `round3_respond` rejects duplicate reveals and commits aggregation atomically. Dealer/DKG bind `(t,n)` and session SSID into randomness; resharing acceptance hashes bind `active_set` (v2); partial PK combine validates coefficients and non-canonical subset masks. Several secret-bearing or panic-prone helpers are narrowed or removed from the public API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f73e713ebe7922b6d824d881c8dab9fbd9a472b0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dilithium/src/ml_dsa_87.rs
+++ b/dilithium/src/ml_dsa_87.rs
@@ -158,7 +158,10 @@ impl fmt::Debug for Keypair {
 /// `Clone` is intentionally not derived because the underlying bytes are sensitive.
 /// To explicitly copy a secret key, use `SecretKey::from_bytes(&sk.to_bytes())?`,
 /// which makes the duplication of secret material visible at every call site.
-#[derive(ZeroizeOnDrop)]
+///
+/// `Zeroize` is derived (in addition to `ZeroizeOnDrop`) so wrapper types that
+/// embed a `SecretKey` can themselves `#[derive(Zeroize, ZeroizeOnDrop)]`.
+#[derive(Zeroize, ZeroizeOnDrop)]
 pub struct SecretKey {
 	bytes: [u8; SECRETKEYBYTES],
 }
--- a/dilithium/src/ml_dsa_87.rs
+++ b/dilithium/src/ml_dsa_87.rs
@@ -222,11 +222,21 @@ impl PublicKey {
 	///
 	/// Returns a PublicKey
 	pub fn from_bytes(bytes: &[u8]) -> Result<PublicKey, KeyParsingError> {
-		let result = bytes.try_into();
-		match result {
-			Ok(bytes) => Ok(PublicKey { bytes }),
-			Err(_) => Err(KeyParsingError::BadPublicKey),
+		let bytes: [u8; PUBLICKEYBYTES] =
+			bytes.try_into().map_err(|_| KeyParsingError::BadPublicKey)?;
+
+		// Reject the degenerate all-zero t1 public key. With t1 = 0 the challenge term in the
+		// verification relation vanishes, letting an attacker forge signatures without any
+		// secret key (see `sign::verify`). Honest key generation never produces t1 = 0, so
+		// this only rejects malformed/malicious keys and never a legitimate one.
+		let mut rho = [0u8; params::SEEDBYTES];
+		let mut t1 = crate::polyvec::Polyveck::default();
+		crate::packing::unpack_pk(&mut rho, &mut t1, &bytes);
+		if t1.vec.iter().all(|p| p.coeffs.iter().all(|&c| c == 0)) {
+			return Err(KeyParsingError::BadPublicKey);
 		}
+
+		Ok(PublicKey { bytes })
 	}
 
 	/// Verify a signature for a given message with a public key.
@@ -343,5 +353,24 @@ mod tests {
 		let keys = Keypair::generate(get_random_bytes());
 		let big_msg = vec![0u8; MAX_MESSAGE_SIZE + 1];
 		assert!(!keys.verify(&big_msg, &[0u8; SIGNBYTES], None));
+	}
+
+	// Malicious-key forgery defense: a public key with an all-zero t1 makes verification
+	// independent of the challenge, enabling signature forgery without a secret key.
+	// `from_bytes` must reject such a key so it can never be constructed or stored.
+	#[test]
+	fn from_bytes_rejects_zero_t1_public_key() {
+		use super::{KeyParsingError, PublicKey, PUBLICKEYBYTES};
+
+		// Arbitrary rho, all-zero t1 region.
+		let mut pk = [0u8; PUBLICKEYBYTES];
+		pk[..crate::params::SEEDBYTES].copy_from_slice(&[0x42u8; crate::params::SEEDBYTES]);
+
+		assert!(matches!(PublicKey::from_bytes(&pk), Err(KeyParsingError::BadPublicKey)));
+
+		// A genuine public key must still round-trip through from_bytes.
+		let keys = Keypair::generate(get_random_bytes());
+		let good = keys.public.to_bytes();
+		assert!(PublicKey::from_bytes(&good).is_ok());
 	}
 }
--- a/dilithium/src/ml_dsa_87.rs
+++ b/dilithium/src/ml_dsa_87.rs
@@ -68,6 +68,19 @@ impl Keypair {
 	/// * 'bytes' - private and public keys bytes
 	///
 	/// Returns a Keypair
+	///
+	/// # Consistency check
+	///
+	/// The public half is re-derived from the secret half and must match the
+	/// supplied public-key bytes exactly; otherwise this returns
+	/// [`KeyParsingError::BadKeypair`]. This prevents importing a keypair whose
+	/// public key does not correspond to its secret key — which would otherwise
+	/// let an object sign with one key while advertising an unrelated public key
+	/// (e.g. a receive address the victim cannot spend from).
+	///
+	/// Note: the `secret` and `public` fields are public, so callers can still
+	/// construct or mutate a `Keypair` with mismatched halves directly. This
+	/// check only guards the deserialization/import path.
 	pub fn from_bytes(bytes: &[u8]) -> Result<Keypair, KeyParsingError> {
 		if bytes.len() != SECRETKEYBYTES + PUBLICKEYBYTES {
 			return Err(KeyParsingError::BadKeypair);
@@ -77,6 +90,15 @@ impl Keypair {
 			SecretKey::from_bytes(secret_bytes).map_err(|_| KeyParsingError::BadKeypair)?;
 		let public =
 			PublicKey::from_bytes(public_bytes).map_err(|_| KeyParsingError::BadKeypair)?;
+
+		// Enforce the cross-field invariant: the public key must be the one that
+		// corresponds to the secret key. The derived pk is public data, so a
+		// non-constant-time comparison is fine.
+		let derived_public = crate::sign::public_key_from_secret(&secret.bytes);
+		if derived_public != public.bytes {
+			return Err(KeyParsingError::BadKeypair);
+		}
+
 		Ok(Keypair { secret, public })
 	}
 
@@ -347,6 +369,31 @@ mod tests {
 		let keys = Keypair::generate(get_random_bytes());
 		let big_msg = vec![0u8; MAX_MESSAGE_SIZE + 1];
 		assert!(!keys.verify(&big_msg, &[0u8; SIGNBYTES], None));
+	}
+
+	// A keypair blob whose public half does not correspond to its secret half must
+	// be rejected. Otherwise an imported keypair could sign with one key while
+	// advertising an unrelated public key (e.g. an unspendable receive address).
+	#[test]
+	fn from_bytes_rejects_mismatched_public_key() {
+		use super::{Keypair, KeyParsingError, KEYPAIRBYTES, SECRETKEYBYTES};
+
+		let keys_a = Keypair::generate(get_random_bytes());
+		let keys_b = Keypair::generate(get_random_bytes());
+
+		// Genuine keypair bytes must round-trip.
+		let good = keys_a.to_bytes();
+		assert!(Keypair::from_bytes(&good).is_ok(), "honest keypair must be accepted");
+
+		// Splice A's secret key with B's (unrelated) public key.
+		let mut forged = [0u8; KEYPAIRBYTES];
+		forged[..SECRETKEYBYTES].copy_from_slice(&keys_a.secret.to_bytes());
+		forged[SECRETKEYBYTES..].copy_from_slice(&keys_b.public.to_bytes());
+
+		assert!(
+			matches!(Keypair::from_bytes(&forged), Err(KeyParsingError::BadKeypair)),
+			"public key not derived from the secret key must be rejected"
+		);
 	}
 
 	// Malicious-key forgery defense: a public key with an all-zero t1 makes verification
--- a/dilithium/src/ml_dsa_87.rs
+++ b/dilithium/src/ml_dsa_87.rs
@@ -387,7 +387,7 @@ mod tests {
 	// advertising an unrelated public key (e.g. an unspendable receive address).
 	#[test]
 	fn from_bytes_rejects_mismatched_public_key() {
-		use super::{Keypair, KeyParsingError, KEYPAIRBYTES, SECRETKEYBYTES};
+		use super::{KeyParsingError, Keypair, KEYPAIRBYTES, SECRETKEYBYTES};
 
 		let keys_a = Keypair::generate(get_random_bytes());
 		let keys_b = Keypair::generate(get_random_bytes());
@@ -414,7 +414,7 @@ mod tests {
 	// public key. `from_bytes` must reject such blobs at import.
 	#[test]
 	fn from_bytes_rejects_corrupted_tr_or_t0() {
-		use super::{Keypair, KeyParsingError, SECRETKEYBYTES};
+		use super::{KeyParsingError, Keypair, SECRETKEYBYTES};
 		use crate::params::{POLYT0_PACKEDBYTES, SEEDBYTES, TR_BYTES};
 
 		let keys = Keypair::generate(get_random_bytes());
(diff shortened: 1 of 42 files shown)


dilithium/src/ml_dsa_87.rs
```diff
@@ -78,6 +90,15 @@ impl Keypair {
 			SecretKey::from_bytes(secret_bytes).map_err(|_| KeyParsingError::BadKeypair)?;
 		let public =
 			PublicKey::from_bytes(public_bytes).map_err(|_| KeyParsingError::BadKeypair)?;
+
+		// Enforce the cross-field invariant: the public key must be the one that
+		// corresponds to the secret key. The derived pk is public data, so a
+		// non-constant-time comparison is fine.
+		let derived_public = crate::sign::public_key_from_secret(&secret.bytes);
```
n13: **[suggestion]** `Keypair::from_bytes` now re-derives the public key from `(rho, s1, s2)` and rejects a mismatched public half — good for the “sign with A, advertise B” class. It still does not check that the secret key’s stored `tr` equals `H(pk)` or that stored `t0` matches the re-derived low bits. A blob with correct `rho/s1/s2/pk` but corrupted `tr` or `t0` is accepted; signing then uses the bad `tr`/`t0`, producing signatures that fail under the advertised public key (or take a weird rejection path). That is mostly self-sabotage / footgun rather than forgery, but the new check is already paying for a full re-derivation, so verifying `tr` and `t0` would close the remaining packed-SK invariant gaps.

**Suggestion:** In `public_key_from_secret` (or at the call site), also compare re-derived `t0` to the unpacked `t0` and `tr` to `SHAKE256(pk)`. Return `BadKeypair` on either mismatch; add a regression test that mutates only the `tr` or `t0` region of an honest SK||PK blob.

threshold/src/keygen/dkg/types.rs
```diff
@@ -47,7 +47,19 @@ use qp_rusty_crystals_dilithium::fips202;
...
+/// `TranscriptSigner` requires [`ZeroizeOnDrop`] because implementors hold a
+/// long-term signing key (`my_signer` in [`DkgConfig`]). The DKG/resharing
+/// state machines own the signer for the whole protocol lifetime and rely on
+/// dropping the configuration to erase that key at their zeroization boundary.
+/// Making the guarantee part of the trait — rather than an unenforced integrator
+/// convention — ensures a signer holding raw secret-key bytes cannot leave them
+/// in freed process memory after the protocol completes or aborts. Implementors
+/// that wrap already-zeroizing key types (e.g. dilithium's `SecretKey`) can
+/// satisfy this with `#[derive(Zeroize, ZeroizeOnDrop)]` or a marker impl.
+pub trait TranscriptSigner: ZeroizeOnDrop {
```
n13: **[suggestion]** `TranscriptSigner: ZeroizeOnDrop` is intended to make long-term auth-key erasure part of the protocol teardown contract. `ZeroizeOnDrop` is only a marker trait: an empty `impl ZeroizeOnDrop for MySigner {}` (as the in-tree `DilithiumSigner` test helper does) satisfies the bound without implementing `Drop`/`Zeroize` and without wiping any key bytes. Integrators can therefore believe the type system guarantees erasure when it does not.

**Suggestion:** Require `Zeroize + ZeroizeOnDrop` (or a custom sealed helper with a documented wipe method), and update docs to state that a bare marker impl is insufficient unless every secret field already implements `ZeroizeOnDrop`. Prefer `#[derive(Zeroize, ZeroizeOnDrop)]` in examples instead of a marker-only impl.

threshold/src/resharing/types.rs
```diff
@@ -714,6 +715,52 @@ pub struct ResharingCertificate {
...
+			let party_id = ParticipantId::deserialize_reader(reader)?;
+			let sig_len = u32::deserialize_reader(reader)? as usize;
+			if sig_len > MAX_ACCEPT_SIGNATURE_LEN {
+				return Err(borsh::io::Error::new(
+					borsh::io::ErrorKind::InvalidData,
+					"ResharingCertificate accept signature exceeds MAX_ACCEPT_SIGNATURE_LEN",
+				));
+			}
+			// Read incrementally rather than pre-allocating the claimed length.
+			let signature = crate::broadcast::read_length_prefixed(reader, sig_len)?;
```
n13, line 758: **[suggestion]** `ResharingCertificate` accept signatures correctly use `read_length_prefixed` so a truncated body cannot force a full up-front allocation of the claimed length. The parallel wire type `ResharingAccept` (same `MAX_ACCEPT_SIGNATURE_LEN` = 8 KiB) is unchanged and still does `vec![0u8; sig_len]` before `read_exact`. The bound is smaller than multi-MB Round 2/3 payloads, but this is the same DoS class the PR is closing elsewhere and is easy to leave as a residual hole on the live accept-message path.

**Suggestion:** Deserialize `ResharingAccept.signature` via `crate::broadcast::read_length_prefixed` the same way certificate accepts do, and add a truncated-body unit test mirroring the Round2/certificate cases.

threshold/src/signer.rs
```diff
@@ -504,16 +513,46 @@ impl ThresholdSigner {
...
+			}
+			for other in round2_data.active_participants.others(me) {
+				if !seen_parties.contains(&other) {
+					return Err(ThresholdError::MissingBroadcast { party_id: other });
+				}
+			}
+
+			// Build the aggregate locally (temporary validated state); commit only on
+			// full success so a failed unpack cannot leave partially-mutated state.
+			let mut aggregated = round2_data.w_aggregated.clone();
```
n13: **[suggestion]** Atomic aggregation correctly builds `w_aggregated` in a local buffer so a mid-stream unpack failure cannot partially mutate state — good. The implementation clones the entire existing aggregate first. For large `k_iterations` (e.g. (4,6) with K=1600) that roughly doubles peak memory during Round 3 (order of tens of MB of `Polyveck`s) on every `round3_respond` call.

**Suggestion:** Prefer unpack-all-then-aggregate without cloning the baseline: e.g. collect/validate all peer `w_other` vectors first, or fold into a fresh buffer seeded once, while keeping the “commit only on full success” property.

threshold/src/signer.rs
```diff
@@ -525,9 +564,16 @@ impl ThresholdSigner {
...
-					aggregate_commitments_dilithium(&mut round2_data.w_aggregated[k_idx], &w_other);
+					aggregate_commitments_dilithium(agg, &w_other);
 				}
 			}
+
+			// All reveals validated and unpacked: commit the aggregate atomically.
+			self.state
+				.round2_data
+				.as_mut()
+				.expect("round2_data present in AfterRound2")
```
n13: **[nit]** After validating `round2_data` with `as_ref().ok_or(...)`, the commit path uses `.as_mut().expect("round2_data present in AfterRound2")`. Under the current control flow this should be unreachable, but `expect` turns a logic bug into a process abort on a public signing API. Elsewhere this PR correctly prefers `Result` over panic for untrusted/setup paths (`run_local_dkg`, length checks, etc.).

**Suggestion:** Reuse `ok_or(ThresholdError::InvalidState { ... })?` for the `as_mut()` commit, matching the earlier check.

hdwallet/src/lib.rs
```diff
@@ -98,17 +99,41 @@ pub fn mnemonic_to_seed(
...
+/// BIP39 requires NFKD Unicode normalization of both the mnemonic and the passphrase
+/// before PBKDF2. The bip39 crate's `parse_in_normalized`/`to_seed_normalized` APIs
+/// assume the *caller* already normalized their inputs, so we normalize here. Without
+/// this, canonically equivalent inputs (e.g. a composed "é" vs a decomposed "e"+combining
+/// accent in a passphrase) would silently derive different seeds and thus different keys.
 fn parse_mnemonic_to_seed(
 	mnemonic: &str,
 	passphrase: Option<&str>,
 ) -> Result<[u8; 64], HDLatticeError> {
+	let normalized_mnemonic = nfkd_owned(mnemonic);
```
n13, line 127: **[suggestion]** Applying BIP39 NFKD to mnemonic and passphrase is the right compliance fix and is well tested for composed/decomposed passphrases and NBSP mnemonics. It is also a silent seed change for any existing wallet whose passphrase or mnemonic was stored/used in a non-NFKD form: after upgrade, `mnemonic_to_seed` / `derive_*_from_mnemonic` will derive different keys than historical tooling that skipped normalization.

**Suggestion:** Call this out explicitly in release notes / breaking-changes (not only as “correctness”), and document a recovery path if any production users may have non-ASCII passphrases. Consider a temporary feature flag or migration helper only if you know such wallets exist.

threshold/src/keygen/dealer.rs
```diff
@@ -263,9 +280,26 @@ fn generate_threshold_shares(
...
+		// mask (and threshold config), rather than assigning raw XOF output to
+		// masks purely by enumeration order. This binds each secret share to the
+		// subset it belongs to, so the derivation cannot be reinterpreted for a
+		// different subset/policy that happens to squeeze the stream in the same
+		// order. `raw` keys the derivation; `honest_signers` is the subset mask.
+		let mut raw = [0u8; 64];
+		fips202::shake256_squeeze(&mut raw, state);
+
+		let mut sh = fips202::KeccakState::default();
+		fips202::shake256_absorb(&mut sh, DEALER_SUBSHARE_DOMAIN);
```
n13, line 293: **[suggestion]** Binding `(t, n)` into the dealer XOF and domain-separating per-subset seeds are real fixes for policy-collision and subset-reinterpretation. Both changes also rewrite the deterministic seed→key map for **all** `generate_with_dealer` outputs, including identical `(t, n, seed)` inputs (not only cross-policy collisions). Any integrator that regenerates shares from a fixed seed will get a different public key and shares after this PR.

**Suggestion:** Document as a hard breaking change for deterministic dealer keygen (separate from “honest callers see no behavior change”). If any environment relies on seed-stable dealer keys, provide a versioned derivation path or migration note (`threshold-dealer-subshare-v1` already versions the PRF).

threshold/src/signer.rs
```diff
@@ -504,16 +513,46 @@ impl ThresholdSigner {
...
 					current: "AfterRound2", // phase already validated above
 					expected: "AfterRound2",
 				})?;
 
+			// Exactly one reveal per other participant recorded at Round 2.
+			let expected_others = round2_data.active_participants.len().saturating_sub(1);
+			if other_round2.len() != expected_others {
+				return Err(ThresholdError::WrongPartyCount {
+					provided: other_round2.len() + 1,
+					required: round2_data.active_participants.len() as u32,
```
n13: **[nit]** When the reveal set size is wrong, the code returns `ThresholdError::WrongPartyCount { provided: other_round2.len() + 1, required: active_participants.len() as u32 }`. `Display` for that variant always says “requires exactly {} (threshold)”. Here `required` is the session participant count (equal to threshold under the protocol’s “exactly T signers” rule, but the field name/message are overloaded). Slightly confusing for operators debugging incomplete/extra reveals.

**Suggestion:** Either introduce a dedicated error (e.g. `IncompleteRevealSet` / `UnexpectedRevealer`) or adjust the `WrongPartyCount` display text to say “signing set size” rather than hard-coding “threshold”.

n13 (approved):
Please check if any of these suggestions make sense - grok 4.5

Approved with comments

PR #96, opened 2026-07-14
Security review again

illuzen:
# Security review remediations: hdwallet, dilithium, threshold

This PR addresses a batch of security review findings across the `hdwallet`, `dilithium`, and `threshold` crates. Each finding was handled the same way: verify the reported behavior against the code, reproduce it with a test that fails on the unfixed code (where the finding is a code defect), fix it, and land the fix with its regression test in a dedicated commit.

### Excluded old members could abort resharing (`95cf4f9`)

Round 5 of the resharing protocol used inconsistent trust boundaries: completion (`have_all_round5`) and the transcript hash only consider the required senders (active set ∪ new committee), but the Combining failure scan honored `success = false` broadcasts from *any* participant. A single old member excluded from the active set — exactly the party the protocol promises to proceed without — could abort every resharing session with one forged failure report, and parties could diverge based on whether they received it before combining.

The required Round 5 sender set is now computed by a single helper used by all three consumers. No legitimate signal is lost: honest failure reports only ever originate from new committee members, which are always required. Covered by a unit test pair (excluded member's failure ignored / active member's failure still aborts) and an end-to-end test where an excluded member withholds its Ready signal and injects a forged failure.

### Signing protocol identity could diverge from the share position (`6f3c94d`)

`DilithiumSignProtocol::new` took a caller-supplied `my_participant_id` but never checked it against `signer.party_id()`. The state machine keyed self-broadcasts under the caller's id while payloads and self-exclusion filters used the share's id; a divergent pair made every peer silently drop the party's messages (payload `party_id` ≠ envelope sender), stalling the session, and desynchronized the RSS share position used by `combine`.

**Breaking change:** the `my_participant_id` parameter is removed entirely. The party's protocol identity is now `signer.party_id()` by construction, making the mismatch unrepresentable. Construction still rejects a signer share whose id is outside the agreed signing set.

### Resharing certificate was not self-describing (`6260bce`)

`ResharingCertificate::verify` treated the caller-supplied `new_participants` slice as the complete set of required acceptors, so an empty or truncated list vacuously verified an unsigned certificate. The new committee is now stored in the certificate and bound into the acceptance hash (domain bumped to `resharing-accept-v3`), and verification demands the trusted key map equal the committee exactly — empty, truncated, superset, and rewritten committees all fail closed.

### Transcript signer zeroization was implicit (`ff3f0fd`)

`DkgState::zeroize()` and `ResharingProtocol::zeroize_session_secrets` relied on drop side effects to erase the signer's long-term key, but `ZeroizeOnDrop` is only a marker trait — a bare-marker impl satisfies the bounds without wiping anything. The signer's `zeroize()` is now called directly, making key erasure a state-machine invariant instead of downstream implementer discipline.

### `Poly` coefficient access sealed behind a validated API (`27c7055`)

The public `poly` module exposed raw `i32` coefficients with comment-only range preconditions. Safe callers could trigger `i32` overflow in `reduce`/`ntt`/`invntt_tomont` (panic in checked builds, silent wraparound in release), and `check_norm` returned a *wrong result* for out-of-range inputs like `i32::MIN`.

- `check_norm` is now total for all `i32` (branchless `i64` absolute value).
- `Poly.coeffs` is `pub(crate)`; external construction goes through the validated `Poly::from_coeffs`, with `coeffs()`/`coeffs_mut()` accessors carrying a documented low-level contract.
- Per-function coefficient bounds are documented and enforced with `debug_assert!` in `reduce`, `ntt`, `invntt_tomont`, and `shiftl`.
- The new assertions surfaced a real latent bug: `compute_partial_pk_t` in `threshold` (and the NTT roundtrip tests) fed `8·Q`-bounded forward-NTT output into an inverse transform that requires `|c| < Q`. A `reduce` step now sits between them, matching the reference implementation.

**Breaking change:** downstream code that accessed `Poly.coeffs` directly must migrate to `from_coeffs`/`coeffs()`/`coeffs_mut()` (the `threshold` crate and examples are migrated in this PR).

### `SecretKey::from_bytes` now validates key invariants (`f2da599`)

`SecretKey::from_bytes` accepted any length-correct byte string, while `Keypair::from_bytes` validated `t0`/`tr` consistency. A corrupted standalone secret key could silently produce signatures that fail verification. Both import paths now enforce the same invariants.

### Removed aliasing `ChildNumber::hardened_from_u32` (`c98139b`)

The numeric constructor OR-ed the hardened bit into the caller-supplied index without rejecting indexes that already had it set, so `hardened_from_u32(1 << 31)` and `hardened_from_u32(0)` encoded identically — and child derivation feeds only those bytes into the HMAC, deriving the same wallet key for two distinct advertised indexes. Distinct account/address indexes could therefore collapse onto one private key.

**Breaking change:** the constructor is removed (nothing in the workspace used it). The string parser, which already enforces `index < 2^31`, is the sole validated construction path; a test pins rejection of hardened-bit aliases at the boundary (`2147483648'` must not alias `0'`).

### Input bounds before expensive work (`0fe405e`, `8607447`)

- Mnemonic and passphrase sizes are capped (1 KiB each) before NFKD normalization and PBKDF2, mirroring the existing derivation-path bounds, so oversized inputs can no longer drive unbounded allocation and CPU work before rejection.
- Derivation-path (and wormhole chain-ID) validation now runs *before* BIP39 seed stretching, so requests guaranteed to fail no longer cost full PBKDF2 work.

### False determinism claim in key derivation (`412622f`)

The `derivation` module claimed "the derived public key is deterministic for the same (master_key, tweak)". It is not: every DKG run mixes a mandatory fresh `session_nonce` into Round 1 randomness (a deliberate anti-grinding property), so re-running the DKG yields a different key. An integrator trusting the stated determinism during re-provisioning or disaster recovery would silently lose the canonical key and everything bound to it.

The docs now state the actual guarantee — one canonical *stored* key per `(master_key, tweak)`, never recomputable — and a new test pins the nonce dependence of the DKG output so docs and code cannot drift apart again. This finding was fixed doc-side because making the key nonce-independent would reintroduce the randomness-grinding attack the nonce binding exists to prevent.

### Misc (`99c72e3`, `feb36ca`, `5e4cf6a`)

`SendMany` transport-requirement docs, rustfmt fixups.

## Compatibility notes

Three public API breaking changes, all of which convert silent misuse into compile errors or remove the unsafe surface outright:

| Crate | Change | Migration |
|---|---|---|
| `threshold` | `DilithiumSignProtocol::new` loses `my_participant_id` | Delete the argument; identity comes from the signer share |
| `dilithium` | `Poly.coeffs` no longer public | Use `Poly::from_coeffs` / `coeffs()` / `coeffs_mut()` |
| `hdwallet` | `ChildNumber::hardened_from_u32` removed | Parse the canonical string form, e.g. `"44'".parse::<ChildNumber>()` |

`ResharingCertificate` from `6260bce` bumps the acceptance-hash domain to `resharing-accept-v3`; certificates produced by older code will not verify against the new format (and vice versa).

## Testing

- Every code fix ships with a regression test that was verified to fail against the unfixed code.
- Full workspace test suite passes (`dilithium`, `hdwallet`, `threshold` unit + integration suites, including the ~5-minute resharing suite).
- `cargo clippy --all-targets` clean; formatted with the project's nightly `rustfmt` configuration.

---

> [!NOTE]
> **High Risk**
> Touches threshold resharing/signing liveness, certificate verification, and secret-key import paths, plus breaking public API changes in dilithium and hdwallet that integrators must migrate.
> 
> **Overview**
> Security review remediations across **hdwallet**, **dilithium**, and **threshold**: protocol liveness/auth fixes, stricter key and polynomial APIs, wallet input bounds, and corrected derivation docs.
> 
> **Threshold:** Resharing Round 5 now uses one `required_round5_senders` set for completion, transcript hashing, and failure aborts, so an old member excluded from the active set cannot stall every session with a forged `success = false`. `ResharingCertificate` stores `new_committee`, bumps acceptance to `resharing-accept-v3`, and verifies against an exact key map (no vacuous empty-committee pass). `DilithiumSignProtocol::new` drops `my_participant_id`; identity is `signer.party_id()`. DKG/resharing state machines call `TranscriptSigner::zeroize()` explicitly instead of relying on `Drop`.
> 
> **Dilithium:** `Poly.coeffs` is sealed behind `from_coeffs` / `coeffs()` / `coeffs_mut()` with documented bounds and `debug_assert!`; `check_norm` is fixed for all `i32`. `SecretKey::from_bytes` validates `tr`/`t0` like `Keypair::from_bytes`. Threshold partial-PK / NTT paths add `reduce` before inverse NTT where forward output can reach `8·Q`.
> 
> **hdwallet:** Removes aliasing `ChildNumber::hardened_from_u32`; caps mnemonic/passphrase at 1 KiB before NFKD/PBKDF2; validates derivation (and wormhole) paths before seed stretching.
> 
> **Breaking:** `DilithiumSignProtocol::new` arity, `Poly.coeffs` visibility, `hardened_from_u32` removal, older resharing certificates vs v3.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f3c94d8a5c33ad24c008c4200b226fd2d35940f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/threshold/src/resharing/protocol.rs
+++ b/threshold/src/resharing/protocol.rs
@@ -1464,12 +1464,16 @@ impl<S: TranscriptSigner> ResharingProtocol<S> {
 	/// members (status) plus new committee members (new-share commitments +
 	/// partial PKs). `None` until the active set is agreed.
 	///
-	/// This is the single trust boundary for Round 5 consumption. Completion
+	/// This is the outer trust boundary for Round 5 consumption. Completion
 	/// ([`Self::have_all_round5`]), the failure-abort scan in Combining, and
-	/// the transcript hash all use exactly this set, so a broadcast from an
-	/// old member excluded from `Act` can never influence the outcome — the
-	/// protocol's liveness promise is that it proceeds without such members,
-	/// and honoring their failure reports would let a single excluded (e.g.
+	/// the transcript hash use exactly this set; the Combining share-data
+	/// checks ([`Self::verify_new_share_consistency`],
+	/// [`Self::verify_public_key_preservation`]) restrict further to new
+	/// committee members, a subset of this set. Together these ensure a
+	/// broadcast from an old member excluded from `Act` never influences the
+	/// outcome — the protocol's liveness promise is that it proceeds without
+	/// such members, and honoring their failure reports (or hard-failing on
+	/// their poisoned share data) would let a single excluded (e.g.
 	/// compromised, being-rotated-out) member abort every session.
 	fn required_round5_senders(&self) -> Option<alloc::collections::BTreeSet<ParticipantId>> {
 		let act = self.active_set.as_ref()?;
@@ -2117,11 +2121,21 @@ impl<S: TranscriptSigner> ResharingProtocol<S> {
 	/// All members of new subset J must produce identical `s_J^new` (and thus identical
 	/// commitments). Cross-verify that.
 	///
-	/// Only accepts share commitments from parties that are actually in the new subset.
+	/// Only considers broadcasts from new committee members (the only honest
+	/// producers of share commitments), and within those, only commitments
+	/// for subsets the sender belongs to.
 	fn verify_new_share_consistency(&self) -> Result<(), ResharingProtocolError> {
 		let mut by_subset: BTreeMap<SubsetMask, Vec<(ParticipantId, [u8; COMMITMENT_HASH_SIZE])>> =
 			BTreeMap::new();
 		for (party, broadcast) in &self.round5_broadcasts {
+			// Round 5 broadcasts are buffered from any participant (an
+			// observer catching up may store them before Act is known), but
+			// only new committee members contribute share commitments. Skip
+			// everyone else before any content check, so a non-required
+			// sender's broadcast cannot influence the outcome.
+			if !self.config.new_participants().contains(*party) {
+				continue;
+			}
 			for (j_mask, commit) in &broadcast.share_commitments {
 				// Only accept commitments from parties that are in this new subset
 				if self.config.new_participants().is_in_mask(*party, *j_mask) {
@@ -2181,7 +2195,9 @@ impl<S: TranscriptSigner> ResharingProtocol<S> {
 	/// Cross-check the broadcast partial PKs and sum them to confirm the
 	/// resharing reconstructs the original public key.
 	///
-	/// Only accepts partial PK contributions from parties that are actually in the new subset.
+	/// Only considers broadcasts from new committee members (the only honest
+	/// producers of partial PKs), and within those, only contributions for
+	/// subsets the sender belongs to.
 	fn verify_public_key_preservation(&self) -> Result<(), ResharingProtocolError> {
 		// The public key is the sum of exactly one partial PK per canonical new subset
 		// (`new_subset_order`). Any other `j_mask` is not part of that sum, so we must
@@ -2194,6 +2210,22 @@ impl<S: TranscriptSigner> ResharingProtocol<S> {
 
 		let mut canonical: BTreeMap<SubsetMask, [[i32; N as usize]; K]> = BTreeMap::new();
 		for (party, broadcast) in &self.round5_broadcasts {
+			// Skip broadcasts from senders outside the new committee BEFORE
+			// the non-canonical-mask hard reject below. Round 5 broadcasts
+			// are buffered from any participant, so without this filter an
+			// old member excluded from the active set — whose broadcast the
+			// session must be able to proceed without — could abort every
+			// session by sending success=true with a single poisoned mask,
+			// and only on the parties that happened to receive it.
+			if !self.config.new_participants().contains(*party) {
+				if !broadcast.partial_pks.is_empty() {
+					log::warn!(
+						"Ignoring partial PKs from party {}: not a new committee member",
+						party
+					);
+				}
+				continue;
+			}
 			for (j_mask, t_partial) in &broadcast.partial_pks {
 				// Reject partial PKs keyed by any mask that is not a canonical new subset.
 				// An honest party never broadcasts one; a malicious party could use it to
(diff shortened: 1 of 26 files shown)


threshold/src/resharing/protocol.rs
```diff
@@ -1452,14 +1460,27 @@ impl<S: TranscriptSigner> ResharingProtocol<S> {
...
 	}
 
+	/// The Round 5 senders this session actually depends on: active old
+	/// members (status) plus new committee members (new-share commitments +
+	/// partial PKs). `None` until the active set is agreed.
+	///
+	/// This is the single trust boundary for Round 5 consumption. Completion
+	/// ([`Self::have_all_round5`]), the failure-abort scan in Combining, and
+	/// the transcript hash all use exactly this set, so a broadcast from an
+	/// old member excluded from `Act` can never influence the outcome — the
```
n13: **[bug]** The new `required_round5_senders` contract claims that a Round 5 broadcast from an old member excluded from `Act` “can never influence the outcome,” and `have_all_round5` / the Combining failure scan / the transcript hash do use that set. That claim is incomplete: `handle_combining` still calls `verify_public_key_preservation()`, which iterates **all** `round5_broadcasts` and **hard-fails** on any non-canonical `partial_pks` mask *before* the `is_in_mask` ignore path. An excluded (or otherwise non-required) old member can still abort every session by sending a syntactically valid Round 5 with `success = true` and a single non-canonical mask (e.g. the full-committee mask). Parties that never received that broadcast proceed, so this also reintroduces the same honest-party divergence hazard the failure-scan fix documents. The new unit/integration tests only forge `success = false` with empty `partial_pks`, so they do not catch this residual DoS.

**Suggestion:** Restrict every Combining consumer of Round 5 data to `required_round5_senders()` (or, for partial PKs specifically, to `new_participants()` only—honest old members never contribute `partial_pks`). Apply the non-canonical-mask hard reject only after that filter. Extend `test_excluded_old_member_cannot_abort_with_forged_round5_failure` (and the Combining unit harness) to forge a non-canonical `partial_pks` entry from the excluded party and assert the session still completes. Soften the doc comment so it matches the actual trust boundary.

threshold/src/resharing/protocol.rs
```diff
@@ -1452,14 +1460,27 @@ impl<S: TranscriptSigner> ResharingProtocol<S> {
...
 	fn have_all_round5(&self) -> bool {
-		// Round 5 has contributions from active old members (status) and new
-		// committee members (new-share commitments + partial PKs). Old members
-		// outside the active set may be offline, so they are not required —
-		// though if an online one reports failure, Combining still honors it.
-		let Some(act) = &self.active_set else { return false };
-		let required: alloc::collections::BTreeSet<ParticipantId> =
-			act.iter().copied().chain(self.config.new_participants().iter()).collect();
+		// Old members outside the active set may be offline, so they are not
+		// required; their Round 5 broadcasts (if any) are ignored outright —
```
n13, line 1485: **[suggestion]** The rewritten `have_all_round5` comment says non-active old members’ Round 5 broadcasts are “ignored outright.” That is true for completion gating and (after this PR) the failure-abort scan / transcript hash, but not for share / public-key verification (see Issue 1). Overstating the invariant makes the next security review or maintainer more likely to miss residual consumers of `round5_broadcasts`.

**Suggestion:** Phrase the comment as “not required for completion / not included in the transcript / not honored for failure-abort,” and explicitly list any remaining consumers that still scan the full map—or delete those residual consumers so the stronger wording becomes true.

threshold/tests/resharing_tests.rs
```diff
@@ -4123,6 +4132,83 @@ fn test_slow_old_only_member_observes_and_completes() {
...
+	let tamper: TamperFn = Box::new(move |sender, _recipient, data| {
+		if sender != 2 {
+			return data;
+		}
+		let msg: ResharingMessage = match borsh::from_slice(&data) {
+			Ok(m) => m,
+			Err(_) => return data,
+		};
+		if matches!(msg, ResharingMessage::Round1(_)) {
+			let forged = ResharingMessage::Round5(ResharingRound5Broadcast {
```
n13, line 4181: **[suggestion]** The new end-to-end regression `test_excluded_old_member_cannot_abort_with_forged_round5_failure` only tampers Round 1 → Round 5 with `success: false` and empty `share_commitments` / `partial_pks`. That pins the exact pre-fix abort path, but does not exercise other ways an excluded sender’s stored Round 5 can still force `ShareVerificationFailed` (non-canonical masks, and any future scan over the full map). The companion unit tests similarly only vary the `success` bit.

**Suggestion:** Add a second tamper (or unit fixture) where the excluded party reports `success: true` with a non-canonical `partial_pks` key and assert completion + usable reshared shares. Treat “any influence from non-required Round 5 senders” as the property under test, not only `ProtocolAborted` from `success = false`.

n13 (changes requested):

n13 (commented):
Verdict: NOT READY TO MERGE yet. This is a comment-only review; I am deliberately not submitting Approve or Request changes.

Blocking findings:

1. Round 5 trust-boundary bypass remains. The new required_round5_senders set is used for completion, the explicit failure scan, and transcript hashing, but verify_public_key_preservation still scans every stored Round 5 broadcast. An excluded old member can therefore send success=true with a non-canonical partial_pks mask and force ShareVerificationFailed. This preserves the same excluded-member DoS and honest-party divergence class the PR is intended to remove. Filter every Round 5 consumer to the required sender set, or limit partial-PK processing to new committee members, and add a regression with an excluded sender carrying malformed/non-canonical partial_pks. This is also captured in the existing inline review discussion.

2. The Poly hardening does not actually close the external safe mutation surface. Poly.coeffs is now crate-private, but pub fn coeffs_mut returns &mut [i32; N] from safe external code. A downstream caller can still write i32::MIN or i32::MAX and then call reduce, ntt, invntt_tomont, or shiftl, reaching the same debug panic/release wraparound behavior described as the vulnerability. A documented low-level contract and debug_assert do not make this a validated API. Please make arbitrary mutation non-safe/non-public, replace it with validated setters/construction, or otherwise expose only operations that preserve the coefficient bounds.

Validation: the PR workspace tests pass in GitHub CI on Ubuntu and macOS, along with format, clippy/docs, no_std, stack-budget, and audit jobs. I could not run cargo locally because this checkout environment has no cargo binary. CodeQL is currently red with two cleartext-logging alerts in hdwallet tests; those appear separate from the two blockers above but should be triaged before merge.

n13 (changes requested):
The latest commit closes the excluded-old-member poisoned-mask path, and the current CI matrix is green. One security blocker remains.

**Blocking — `dilithium/src/poly.rs:100`: the validated `Poly` API is still bypassable from safe external code.** `pub fn coeffs_mut(&mut self) -> &mut [i32; N]` lets any downstream caller write `i32::MAX`/`i32::MIN`, then call safe public operations such as `reduce`, `ntt`, `invntt_tomont`, or `shiftl`. That recreates the debug panic/release overflow behavior this hardening is intended to remove. A documented precondition plus `debug_assert!` does not make a safe public API sound or validated. Please make unchecked mutation non-public (or `unsafe` with an explicit safety contract), and expose validated setters/transformations for external callers. Internal workspace users can use a crate-private mechanism or validated construction. Add an external-facing regression/compile-fail test proving arbitrary safe mutation is unavailable.

I re-reviewed head `f0822e1a`. All GitHub checks currently pass; the blocker is API-level and is not exercised by those tests.

n13 (approved):
Blocking issue to be addressed in follow up PR

PR #97, opened 2026-07-15
Unify NTT

illuzen:
# NTT unification + security-review follow-ups

Two small follow-ups from the last security-review PR, and the removal of a long-standing wart: the threshold crate's duplicate Go/circl-compatible NTT implementation.

## NTT unification (threshold)

The threshold crate carried its own NTT (`circl_ntt.rs`, ported from Cloudflare's circl library) alongside the dilithium crate's FIPS 204-style NTT. The two transforms agree mod Q but use different coefficient representatives — unsigned `[0, 18Q)` Montgomery values vs. signed `~(−8Q, 8Q)` — and the split forced the protocol code to maintain two sets of reduction conventions.

### Why the swap is transcript-safe (`7ba54c1`)

Before touching any call site, an equivalence test pins that the circl pipeline (`ntt` → `mul_hat` → `inv_ntt`) and the dilithium pipeline (`poly::ntt` → `pointwise_montgomery` → `invntt_tomont`) produce identical results mod Q for the forward transform, the full multiply pipeline, and the bare inverse — i.e. their Montgomery-factor accounting is identical and only the representatives differ. A full audit of the protocol confirmed that **no NTT-domain representative is ever hashed or serialized**: commitments, transcripts, broadcasts, and signatures all operate on canonical `[0, Q)` standard-domain values, so swapping representatives cannot change any transcript or wire format. (The test is committed for reviewability and deleted together with the module two commits later.)

### The migration (`1cb828c`)

Every circl call site now uses the dilithium NTT: Round 1's `w = A·y`, Round 3's `c·s1` and `c·s2`, combine's `Az` and `c·2^d·t1`, share recovery, and dealer keygen. Two helpers had latent circl-convention assumptions that had to be generalized:

- **`NttAccumulator`** assumed non-negative circl output and normalized negatives with a single `+Q`. Dilithium's forward NTT emits signed values up to ~8Q in magnitude, which that scheme would have mis-wrapped into huge `u64` values. It now canonicalizes each coefficient to `[0, Q)` before accumulating.
- **`normalize_assuming_le2q`** only tolerated negatives down to −Q. It now accepts any `|c| < 2Q`, with a `debug_assert` on the bound.

A pleasant side effect: the manual unsigned sign-fixup + `reduce_le2q` loops that bracketed every inverse NTT collapse into a single `poly::reduce`, which directly satisfies `invntt_tomont`'s documented `|c| < Q` input contract.

### The deletion (`39f166a`)

`circl_ntt.rs` is removed along with its Go-pinned test vectors (exact 18Q-band representatives only the circl transform produces — historical artifacts, not live interop; cross-implementation compatibility is Rust-to-Rust plus standard ML-DSA-87 verification, both covered by the end-to-end suites).

### Round 5 trust boundary documented per consumer (`4c43164`)

The `have_all_round5` comment claimed non-required senders' Round 5 broadcasts are "ignored outright", which overstates what that function alone guarantees. The comment now spells out the three specific guarantees (not honored by the Combining failure scan, not included in the transcript hash, not read by the share-data checks), and `required_round5_senders` documents that its consumer enumeration is exhaustive and that any new consumer of `round5_broadcasts` must apply the same filtering.

### Non-required Round 5 senders pinned as unable to influence Combining (`fed275a`)

The regression tests previously only forged `success = false`. The end-to-end forgery scenario is now a shared helper run twice: once with the forged failure report plus a poisoned non-canonical partial-PK mask, and once with `success = true` where the poisoned mask is the *only* possible abort vector — isolating the share-data path. Both assert full completion and that the reshared shares produce a valid signature. The Combining unit harness gains an all-success variant exercising the same path. Both new tests were verified to fail against the code without the new-committee sender filter in `verify_public_key_preservation`.

## Testing

- Full workspace suite passes: 558 tests across `dilithium`, `hdwallet`, and `threshold`, including the ~10-minute end-to-end signing and resharing suites, which verify real ML-DSA-87 signatures produced by the migrated NTT path.
- The NTT equivalence test (committed mid-stack, deleted with the module) proves mod-Q agreement of both pipelines in history.
- New forgery tests verified red against unfixed code.
- `cargo clippy --workspace --all-targets` clean; formatted with the project's nightly `rustfmt` configuration.

No wire-format or transcript changes. The only public API change is the removal of `qp_rusty_crystals_threshold::circl_ntt`, which was exposed solely for cross-language NTT testing against the Go reference.

---

> [!NOTE]
> **Medium Risk**
> Touches the ML-DSA threshold signing and keygen NTT path across many call sites; risk is mitigated by mod-Q equivalence rationale and no claimed wire/transcript changes, but a bug here would break signatures or keys.
> 
> **Overview**
> Removes the duplicate **circl** NTT module (`circl_ntt`) and routes all threshold crypto through the **dilithium** crate: `poly::ntt`, `poly::pointwise_montgomery`, and `poly::invntt_tomont` in dealer keygen, share recovery, signing (Rounds 1/3/combine), and NTT dot products. The public **`qp_rusty_crystals_threshold::circl_ntt`** export goes away with it.
> 
> **`NttAccumulator`** and **`normalize_assuming_le2q`** are updated for dilithium’s signed NTT representatives (canonicalize to `[0, Q)` before summing; accept `|c| < 2Q`). Signing no longer does manual `reduce_le2q` around inverse NTT—**`poly::reduce`** plus **`invntt_tomont`** replaces that.
> 
> Resharing follow-ups: clearer docs that only **required Round 5 senders** may affect Combining, plus regression tests (unit + e2e) where an **excluded old member** forges Round 5 (`success = false` or a poisoned non-canonical partial-PK mask) and the session must still complete with valid signatures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39f166ad8a5b4684cc6c690fb0933d9ac4a4d649. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13 (approved):
## Review

Reviewed the full diff plus surrounding code on the branch, with the focus on whether the NTT swap is actually behavior-preserving. I re-verified the bounds arithmetic against the dilithium crate's documented contracts and re-ran the relevant tests locally. **Verdict: approve.**

### What was verified independently

**Montgomery/bounds accounting at every migrated call site.** I traced each new call sequence against the contracts in `dilithium/src/poly.rs` / `reduce.rs`:

- `montgomery_reduce` output is strictly `(-Q, Q)`, so `pointwise_montgomery` products feed `invntt_tomont` directly in Round 3 (`c·s1`, `c·s2`) — contract satisfied with no reduction step, as the new comments claim.
- `compute_ntt_dot_product` sums L = 7 products → `|c| < 7Q`; Round 1 and dealer keygen apply `poly::reduce` (output ±6283008 < Q) before `invntt_tomont`. ✓
- Combine's `Az − 2^d·c·t1` is bounded by `7Q + Q = 8Q`, far inside `reduce`'s `2^31 − 2^22 − 1` precondition. ✓
- Forward-NTT inputs everywhere (γ₁-bounded `y_k`, ±1 challenge, η-bounded shares, centered `z` ≤ Q/2, `t1 << D` < Q) are tiny relative to `ntt`'s `2^31 − 1 − 8Q` limit. ✓
- Round 1's `invntt output (< Q) + e_k`: `HyperballSampleVector::round()` centers `e_k` into `[−Q/2, Q/2]`, so `|c| < 1.5Q < 2Q` — the relaxed `normalize_assuming_le2q` precondition holds. Its other callers (`aggregate_commitments_dilithium`: two canonical values, max `2Q − 2`; Round 3 / combine: `< Q` post-inverse-NTT) also satisfy the strict `< 2Q` debug_assert.
- `NttAccumulator` canonicalization to `[0, Q)` before the u64 sum is mod-Q-equivalent to the old raw-circl-value sum, and its `finalize()` output in `[0, Q)` is a valid signed representative (`|c| < Q`) for every downstream consumer (dealer `pointwise_montgomery` + `invntt_tomont`, signing via `recover_share`).

**The equivalence test does what the PR claims.** Checked out `7ba54c1` and ran it — forward transform, full multiply pipeline, and bare inverse all agree mod Q. Since the whole question of this migration is Montgomery-factor accounting, pinning that in history before the swap (and deleting it with the module) is exactly the right shape for this change.

**Transcript-safety spot-check.** Confirmed the values that get hashed/packed (Round 1 `w` commitments, `z`, `w1`) all pass through `normalize_assuming_le2q` / canonical packing first; NTT-domain representatives stay internal (`recover_share` output, `s1h_total`) and are never serialized.

**Round 5 trust-boundary docs are accurate.** All five consumers of `round5_broadcasts` were checked against the new comments: completion and the failure scan filter through `required_round5_senders`, transcript hashing reads only required parties, and both share-data checks (`verify_new_share_consistency`, `verify_public_key_preservation`) skip non-new-committee senders **before** any content validation, including the non-canonical-mask hard reject. The docs now match the code precisely.

**Tests re-run locally** (debug build, so the new `debug_assert` bounds contracts were live):
- 227 threshold unit tests pass at the PR head, including the new all-success Combining variant.
- Both e2e forgery regression tests pass, exercising full resharing + real ML-DSA-87 signing through the migrated NTT path with debug assertions enabled.
- Equivalence test passes at the mid-stack commit.

### Findings (all minor)

1. **Stale circl-era comment** — `threshold/src/protocol/secret_sharing.rs:277` still says "After NTT, coefficients are bounded by 18*Q", which describes the deleted circl transform; the dilithium bound is ~8Q signed. The equivalent comment in `primitives.rs` was updated; this one was missed.

2. **Semver** — removing `pub mod circl_ntt` is a breaking API change to `qp-rusty-crystals-threshold`, currently at 1.0.0 with no version bump in this PR. Fine if versions are bumped at release time, but worth flagging so it doesn't ship as a patch release.

3. **Observation, no action needed** — `NttAccumulator::add_poly`'s `((c % q) + q) % q` is a modulo over secret share data on the keygen/recovery paths. Since `q` is a compile-time constant, LLVM lowers this to multiply-shift (constant-time), and the code it replaces branched on the secret's sign anyway — so this is not a regression, just noting it given the repo's recent constant-time work.

### Summary

The risky part of this PR — swapping the NTT under a threshold signing protocol — is done the way it should be: equivalence pinned in history before the swap, representative-convention assumptions in `NttAccumulator`/`normalize_assuming_le2q` generalized rather than papered over, and the bounds reasoning written into comments that match the actual contracts in the dilithium crate (I verified each one). Net −243 lines, one NTT convention instead of two, and the manual reduction loops around inverse NTT collapse into documented contract-satisfying calls. The Round 5 test hardening closes the remaining untested path (share-data-only influence from a non-required sender) with tests verified red against unfixed code.

---
*Review verified locally: bounds contracts traced against `dilithium` crate sources; 227 unit tests + both e2e forgery tests + mid-stack equivalence test re-run and passing.*

🤖 Generated with [Claude Code](https://claude.com/claude-code)

n13 (approved):
LGTM - please update the comment (this is why I discourage comments - they get out of sync)

PR #98, opened 2026-07-17
Security audit fixes: input validation, DoS hardening, and API contract enforcement

illuzen:
This PR addresses a batch of security-audit review items across the `dilithium` and `threshold` crates. Every fix follows the same discipline: the issue was first reproduced with a failing regression test, then fixed, and the test now passes. All fixes ship with their regression tests.

### Key material and share validation

- **Keypair seed preimage built in a fixed buffer** (`dilithium`): the keygen seed preimage is assembled in a fixed-size stack buffer instead of a heap allocation, so no copy of the seed material is left behind for the allocator to recycle un-zeroized.
- **All-zero `t1` rejected on every key-import path** (`dilithium`/`threshold`): the all-zero public-key rejection is now enforced consistently across every import path, not just some of them.
- **Share coefficient ranges validated at deserialization** (`threshold`): private key shares deserialized from untrusted bytes now have their coefficient ranges checked up front, rejecting corrupt or malicious share data at the trust boundary.
- **Resharing old committee must match the share's DKG list** (`threshold`): the resharing protocol rejects an old-committee list that does not match the participant set recorded in the share at DKG time.

### Protocol correctness

- **Round 3 reveals bound to the Round 1 hashes seen in Round 2** (`threshold`): `round3_respond` verifies reveals against the commitment hashes frozen at Round 2 reveal time rather than a caller-supplied Round 1 set, preserving the commit-reveal anti-rushing property. A rushing peer can no longer substitute a Round 1 hash chosen after observing honest reveals.
- **Round 4 combination failure returns `ProtocolFailed`** (`threshold`): the leader's combination-failure branch returned `SigningError` while the module docs (and the `Failed`-state behavior of the same instance) promise `ProtocolFailed` — the variant integrations use to dispatch fresh-instance retries. Normal ML-DSA rejection-sampling failures are no longer misreported as permanent signing errors.

### Denial-of-service hardening

- **Message/context validated before Round 2 commitment packing** (`threshold`): `round2_reveal` now enforces the ML-DSA message and context bounds *before* packing `k_iterations × 5888` bytes of commitment data (~9.4 MB for 4-of-6). Previously a guaranteed-to-fail request (e.g. a 256-byte context) still forced the full allocation and serialization. Verified by a regression test using a tracking allocator.
- **Resharing Round 5 `error_message` bounded during deserialization** (`threshold`): the one variable-length resharing field without a length bound. A ~90-byte broadcast claiming a near-`u32::MAX` string length forced a 1 MiB allocation on every recipient (borsh 1.6.1 caps the eager allocation there — the audit's 4 GiB figure does not reproduce at this version), and a fully-delivered payload of up to 4 GiB would have been accepted and retained. Now bounded by `MAX_ERROR_MESSAGE_LEN` (1 KiB) using the same chunked-read path as every other variable-length field, with UTF-8 validation.
- **`SignMessageBuffer` enforces its own participant bound** (`threshold`, breaking): the public out-of-order message buffer documented "at most MAX_PARTIES entries per round" but accepted arbitrary `party_id`s, making it an unbounded storage sink for megabyte-scale payloads when used directly. `SignMessageBuffer::new` now takes the session's `ParticipantList` and drops messages from non-participants; the `Default` impl is removed.

### API contract enforcement (breaking)

- **XOF block-squeeze made type-safe** (`dilithium`, breaking): `shake128_squeezeblocks`/`shake256_squeezeblocks` previously took `(&mut [u8], nblocks)` and silently underfilled a short slice (`chunks_exact_mut(r).take(nblocks)`), leaving stale bytes the caller would treat as XOF output and desynchronizing the Keccak state. The functions now take whole blocks (`&mut [[u8; RATE]]`), so a count/capacity mismatch is unrepresentable. All samplers were migrated; the `off` carry-over logic inherited from the C reference in `poly::uniform` was removed as dead code (SHAKE128's rate is divisible by 3, so it never fired). NIST ACVP known-answer tests confirm byte-identical output.
- **`unpack_polyveck_w` validates buffer length** (`threshold`): an undersized buffer is now a recoverable `Err` instead of an out-of-bounds slice panic (which would abort the process in `panic=abort` deployments).

## Breaking changes

- `qp-rusty-crystals-dilithium`: `shake128_squeezeblocks` / `shake256_squeezeblocks` signatures changed from `(out: &mut [u8], nblocks: usize, state)` to `(out: &mut [[u8; RATE]], state)`. Requires a version bump on next publish.
- `qp-rusty-crystals-threshold`: `SignMessageBuffer::new()` now requires the session `ParticipantList`; `Default` removed.

## Test plan

- [x] Each fix has a dedicated regression test that failed before the fix and passes after (allocation-tracking tests for the two DoS items; equivalence and boundary tests elsewhere)
- [x] Full workspace suite green in release mode: 572 tests, 0 failures
- [x] Dilithium NIST ACVP known-answer vectors pass (confirms the XOF/sampler refactor is byte-identical)
- [x] `no_std` compatibility preserved (no new `std` dependencies outside test code)

---

> [!NOTE]
> **High Risk**
> Touches cryptographic key handling, XOF output correctness, threshold commit-reveal security, and multiple untrusted deserialization boundaries; breaking public API changes require coordinated version bumps.
> 
> **Overview**
> This PR batches **security-audit** fixes in `dilithium` and `threshold`, each paired with a regression test.
> 
> **Dilithium:** `keypair` now builds the SHAKE preimage in a **fixed stack buffer** instead of a `Vec` that could free an uncleared seed copy. **SHAKE block squeeze** APIs take `&mut [[u8; RATE]]` so slice length vs. block count cannot silently desync Keccak state; samplers were updated. **`public_key_from_secret`** and standalone **`SecretKey::from_bytes`** now reject secret keys whose derived public key has **all-zero `t1`**, aligned with verifier/`PublicKey::from_bytes`.
> 
> **Threshold signing:** Round 1 commitment hashes are **frozen at Round 2 reveal**; Round 3 checks reveals against that map (not caller-supplied Round 1), blocking post-hoc commitment substitution. **`round2_reveal`** validates message/context **before** large commitment packing. Round 4 combination failures return **`ProtocolFailed`** for documented retries. **`SignMessageBuffer`** only buffers messages from the session **`ParticipantList`** (breaking: `new(participants)`, no `Default`).
> 
> **Import / protocol hardening:** Threshold **`PublicKey`** paths call canonical ML-DSA parse (forgeable zero-`t1` rejected). **`SecretShareData`** Borsh deserialize checks coefficients in `(-Q, Q)`. Resharing config requires **old committee == share’s embedded DKG list**. Round 5 **`error_message`** bounded at deserialize; **`unpack_polyveck_w`** returns `Err` on short buffers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b54db0ebd51f4cfd20fbfe089ca70bcddf0dab0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,9 +45,9 @@ checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
 
 [[package]]
 name = "ar_archive_writer"
@@ -92,9 +92,9 @@ dependencies = [
 
 [[package]]
 name = "bitcoin_hashes"
-version = "0.14.100"
+version = "0.14.101"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c9901a56e133a1fc86eeb1113e2591f45f4682451ca893bff494d2f88918e3f"
+checksum = "bca4c7abb40c8817d77403c880988cfd484f23ab2365726afb2f798363e2c4a2"
 dependencies = [
  "hex-conservative",
 ]
--- a/dilithium/src/fips202.rs
+++ b/dilithium/src/fips202.rs
@@ -408,12 +408,16 @@ fn keccak_absorb_once(s: &mut [u64; 25], r: usize, input: &[u8], p: u8) {
 	s[(r - 1) / 8] ^= 1u64 << 63;
 }
 
-/// Squeeze step of Keccak. Squeezes full blocks of r bytes each.
+/// Squeeze step of Keccak. Squeezes one full rate-sized block per entry of `out`.
 /// Modifies the state. Can be called multiple times to keep squeezing, i.e., is incremental.
 /// Assumes zero bytes of current block have already been squeezed.
-fn keccak_squeezeblocks(out: &mut [u8], nblocks: usize, s: &mut [u64; 25], r: usize) {
-	// Process exactly nblocks blocks using chunks_exact_mut for safe iteration
-	for block in out.chunks_exact_mut(r).take(nblocks) {
+///
+/// The output is typed as whole blocks (`[[u8; R]]`) rather than a flat slice
+/// plus a separate block count, so a count/capacity mismatch — which would
+/// silently underfill the buffer and desynchronize the state — is
+/// unrepresentable at the type level.
+fn keccak_squeezeblocks<const R: usize>(out: &mut [[u8; R]], s: &mut [u64; 25]) {
+	for block in out.iter_mut() {
 		keccakf1600_statepermute(s);
 		for (i, chunk) in block.chunks_exact_mut(8).enumerate() {
 			// chunks_exact_mut(8) guarantees exactly 8 bytes
@@ -434,11 +438,17 @@ pub fn shake128_finalize(state: &mut KeccakState) {
 	state.pos = SHAKE128_RATE;
 }
 
-/// Squeeze step of SHAKE128 XOF. Squeezes full blocks of SHAKE128_RATE bytes each.
+/// Squeeze step of SHAKE128 XOF. Squeezes one full block of SHAKE128_RATE bytes
+/// per entry of `output`.
 /// Can be called multiple times to keep squeezing.
 /// Assumes new block has not yet been started (state->pos = SHAKE128_RATE).
-pub fn shake128_squeezeblocks(output: &mut [u8], nblocks: usize, s: &mut KeccakState) {
-	keccak_squeezeblocks(output, nblocks, &mut s.s, SHAKE128_RATE);
+///
+/// Taking whole blocks (`[[u8; SHAKE128_RATE]]`) instead of a flat slice plus a
+/// block count makes an inconsistent count/capacity pair unrepresentable: the
+/// old shape silently underfilled a short slice, leaving stale bytes that the
+/// caller would treat as XOF output.
+pub fn shake128_squeezeblocks(output: &mut [[u8; SHAKE128_RATE]], s: &mut KeccakState) {
+	keccak_squeezeblocks(output, &mut s.s);
 }
 
 /// Absorb step of the SHAKE256 XOF; incremental.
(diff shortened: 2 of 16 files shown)


n13 (approved):
## Review — verified locally, not just read

Checked out the branch, read every changed line plus its surrounding context, and ran the full workspace suite in release mode: **586 tests, 17 binaries, all green** (exit 0). CI is also fully green (ubuntu + macOS, clippy, no_std wasm build, thumbv7em stack budget, security audit, CodeQL).

**Verdict: Approve.** Each fix is correct, minimal, and paired with a regression test that actually exercises the original failure mode. Details per item below.

### Key material and share validation

- **Fixed-buffer keypair preimage** (`sign.rs`): correct. `seed_bytes` is a `[u8; 32]` from `SensitiveBytes32::into_bytes`, so the fixed `[u8; SEEDBYTES + 2]` preimage can't length-mismatch, and `preimage.zeroize()` is still called at the end of `keypair` (sign.rs:101). The described old-`Vec` failure mechanism is accurate: `Vec::new() + extend_from_slice(32 bytes)` sizes capacity to exactly 32, so the two `push`es forced a realloc that freed a seed-bearing block outside `zeroize`'s reach. The allocator-scanning regression test is sound (block is still valid inside `dealloc`, one test per binary so nothing races the scanner).
- **All-zero `t1` on every import path**: complete coverage. Both `SecretKey::from_bytes` and `Keypair::from_bytes` route through `sign::public_key_from_secret`, which is where the new `t1_nonzero` check lives — so one check covers all SK import paths, and the crafted `s1 = s2 = 0` regression test proves it fires. Threshold `PublicKey` now delegates to the canonical `ml_dsa_87::PublicKey::from_bytes` parser on both `from_bytes` and Borsh, closing the wrapper bypass.
- **Share coefficient ranges at deserialization** (`keys.rs`): manual `BorshDeserialize` for `SecretShareData` rejects coefficients outside `(-Q, Q)` before they can reach `poly::ntt` (whose bound is caller-enforced — panic in debug, silent wrap in release). Boundary test at exactly `±Q` vs `±(Q-1)` is the right test.
- **Resharing old-committee match** (`resharing/types.rs`): the mask-bit-to-party identity-mapping argument in the comment is exactly right, and `ParticipantList::PartialEq` compares the sorted ID vectors, so the comparison is semantic, not structural. Swapped-members, superset, and exact-match cases all tested.

### Protocol correctness

- **Round 3 anti-rushing** (`signer.rs`): this is the most security-critical change in the batch and it's done properly. The Round 1 hashes are frozen into `Round2Data.round1_commitments` at reveal time, `round3_respond` verifies against that map, and the supplied `other_round1` is retained only as a defense-in-depth consistency check — so the function signature (and existing integrations) are unchanged. The exact-set check still runs against the `active_participants` recorded in Round 2, so the frozen map and the required reveal set can't drift apart. The rushing-attack regression test (placeholder Round 1 hash, genuine pair presented at Round 3 → `CommitmentMismatch`) demonstrates the original hole and the fix. Note the deliberate error-precedence change: `InvalidState` now precedes reveal-validation errors — that's the more correct ordering.
- **Round 4 `ProtocolFailed`**: matches the module docs, the trust model, and the `Failed`-state behavior of the same dead instance. The test (force `Round4Deciding` with empty broadcasts) pins the contract integrations rely on for fresh-instance retries.

### DoS hardening

- **`round2_reveal` validates before packing**: `validate_message`/`validate_context` now run before the `k × 5888`-byte commitment packing; `process_round2` still re-checks as defense in depth. The tracking-allocator test proves a rejected call allocates less than one packed commitment and doesn't consume the session.
- **Round 5 `error_message` bound**: the manual `Option<String>` decode mirrors borsh's wire format exactly (flag byte, u32 length, and even borsh's own invalid-flag error message), the 1 KiB bound is checked *before* reading, the read goes through the same chunked `read_length_prefixed` path as the other variable-length fields (verified: it reads exactly `len` bytes in ≤64 KiB steps given the already-consumed prefix), and UTF-8 is enforced. Serialize side stays derived borsh, so round-trips are symmetric — the test confirms both `Some` and `None`.
- **`SignMessageBuffer` participant bound**: the buffer now enforces its own documented invariant instead of relying on `DilithiumSignProtocol::message` to pre-filter. All construction sites migrated (grep-verified), `Default` removed, and the integration test shows non-participant messages dropped while participants are retained.

### XOF type-safety (breaking)

- The `keccak_squeezeblocks` refactor preserves the permute-then-extract order and is applied uniformly; grep confirms no old-style call sites remain anywhere in the workspace. The `poly::uniform` `off` carry-over removal is safe: SHAKE128's rate (168) is divisible by 3, so `off` was provably always 0 and the `+2` slack bytes were dead — the new block-at-a-time refill is byte-identical, which the ACVP KATs passing confirms. The one-shot `shake256` rewrite (whole blocks + `into_remainder` tail) is equivalent for all output lengths, including sub-rate. Making a count/capacity mismatch unrepresentable is strictly better than the old silent underfill.

### Nits (non-blocking)

- `round3_respond` clones the frozen commitments map each call (≤5 × 32-byte entries) to satisfy the borrow checker before the later `as_mut()` — negligible, not worth restructuring.
- `other_round1` in `round3_respond` is now nearly vestigial; keeping it for API compatibility is the right call, but the docs already say it's defense-in-depth only — fine as-is.
- The breaking changes (`shake*_squeezeblocks` signatures, `SignMessageBuffer::new`/`Default` removal) are accurately documented in the PR body; just make sure the version bumps land at publish time as noted.

Clean, well-scoped security work with real regression tests. LGTM.

PR #104, opened 2026-07-23
Security review round 4: secret hygiene, bounded deserialization, and protocol atomicity

illuzen:
This PR addresses the fourth round of security review findings across the `dilithium`, `hdwallet`, and `threshold` crates. Every fix follows the same discipline: the finding was first reproduced with a failing regression test, then fixed, then re-verified (test red on the old code, green with the fix). 19 commits, 28 files, ~2600 insertions.

## Secret hygiene (zeroization)

**Heap intermediates are wiped before their memory is freed.** A new `threshold/tests/heap_zeroization.rs` suite installs a scanning global allocator that inspects every freed block for the scenario's secret pattern at `dealloc` time. It caught, and now guards, six scenarios:

- `derive_dkg_contribution`'s share-digest linearization buffer, incoming DKG Round 1 private frames carrying K_S, and the hyperball sampling scratch buffer (`7d63cf9`);
- the resharing protocol's `party_key` derivation buffer, Round 4 transport frames (both directions — `Action::SendPrivate` now carries `Zeroizing<Vec<u8>>`), sub-share derivation output, and per-polynomial hashing scratch buffers (`ce8175f`);
- DKG K_S relocation out of Copy-typed containers: `transition_to_round2` freed the `received_shared_secrets` map nodes with the secrets still in them (the state zeroizer runs too late — the field is already `None`), `pending_privates.pop()` left queued `Round1Private` bytes beyond the `Vec`'s shrunken length, and the buffered-privates drain consumed its map the same way (`4e6cb6d`);
- the wormhole Poseidon preimage (`preimage_felts`) was `clear()`ed but never wiped; covered by a dedicated allocator-instrumented test in `hdwallet/tests/wormhole_zeroization.rs` (`7d63cf9`).

**Stack copies of the ML-DSA-87 secret key are wiped in import/serialize paths.** `Keypair::from_bytes`, `SecretKey::from_bytes`, and `Keypair::to_bytes` left plaintext `[u8; SECRETKEYBYTES]` copies in dead stack frames because `[u8; N]` is `Copy`. All three now stage the secret in `Zeroizing` locals. A painted-stack probe test (`dilithium/tests/import_stack_zeroization.rs`, release-mode only — debug builds create compiler temporaries no source fix can wipe) verifies no copy survives (`2197133`).

**Debug output no longer leaks key material.** The resharing `Action::SendPrivate` payload (serialized sub-shares) is redacted from the manual `Debug` impl (`df8dee3`).

## Bounded, validating deserialization

- `ResharingActProposal::active_set` length is bounded at deserialize; a hostile length prefix can no longer drive unbounded allocation (`62f736e`).
- `PrivateKeyShare` metadata (threshold, party counts, participant list, share-map shape) is cross-checked at Borsh import and at `ThresholdSigner` construction (`345c7aa`).
- ML-DSA-87 secret keys whose `s1`/`s2` coefficients decode outside `[-ETA, ETA]` are rejected at import: `eta_unpack` and `unpack_sk` now return a canonicality flag (`#[must_use]`) that `public_key_from_secret` enforces, so non-canonical 3-bit slots (5–7) can no longer cross the boundary (`19e2bf0`).
- `SubsetContribution` (exported, previously derive-deserialized into unbounded `Vec<[i32; N]>`) now has a manual deserializer enforcing exactly `L`/`K` polynomials — length prefix checked before any allocation — and η-bounded coefficients, mirroring `SecretShareData` (`3ec5b15`).
- Invalid `DkgConfig` values are unconstructible: fields are private with accessors, and `all_broadcasts_received` uses total quorum arithmetic that cannot underflow on an empty participant list (`e1df0ca`).

## DoS hardening (no work before validation)

- The signing protocol checks the fixed-header SSID and a **per-config** frame budget (`k_iterations`-derived, ~23 KB for a (2,2) session instead of the global 12 MiB) before deserializing anything (`0428526`).
- `ExtendedPrivKey::derive` parses and validates the derivation path first and bounds the seed to the BIP32 master-seed range (16..=64 bytes, new `Error::InvalidSeedLength`) before doing any HMAC work (`bf9b646`).
- Round 2 reveals are rejected on the exact per-config length **before** the SHAKE256 commitment hash runs, so a peer who pre-committed to an oversized blob (up to the 10.5 MB global deserializer bound) cannot force megabytes of hashing in a session whose legitimate reveal is kilobytes; protocol intake drops hash-bound but mis-sized reveals in O(1) (`36e1079`).

## Protocol state-machine correctness

- Aborted resharing sessions no longer strand the old share: recovery restores it so the party can rejoin a retried session (`5e5c7ef`).
- `round3_respond` failures **after** the peers' reveals were folded into the commitment aggregate (e.g. `recover_share` rejecting an active party outside the DKG set — only detectable at that point) now reset the session instead of leaving a poisoned aggregate in the "cleanly retryable" `AfterRound2` state, where a retry would double-count every peer's commitment. Pre-commit-point failures still leave the session clean for corrected retries (`243bb0d`).
- `DerivedKeyId` is bound to the DKG output public key, preventing cross-key derivation-ID collisions (`5599c69`).

## API contract enforcement

- `poly::shiftl`'s unchecked left shift is no longer reachable through public API with out-of-contract coefficients: it is crate-private and the `k_shiftl` docs state the real `2^(31-D)` bound (`6fe967c`).
- `KeccakState` encodes the SHAKE rate as a const-generic parameter, making the absorb/squeeze phase-and-rate mismatches flagged by review unrepresentable, and squeeze chunking canonical (`225e629`).
- `uniform_eta`'s rejection-sampling retry loop is documented as not a timing channel: two fixed blocks yield 544 nibbles at 15/16 acceptance, so a third block is a < 2^-600 event, and rejection counts are independent of the accepted values under standard XOF assumptions (`ec388fe`).

## Compatibility notes

Behavioral tightenings downstream consumers may notice:

- `hdwallet`: seeds outside 16..=64 bytes are now rejected by `ExtendedPrivKey::derive` (new `Error::InvalidSeedLength` variant). The Quantus SDK's `derive_hd_path` passes 64-byte BIP39 seeds and is unaffected.
- `threshold`: resharing `Action::SendPrivate` now carries `Zeroizing<Vec<u8>>`; `DkgConfig` fields are private (use the accessors); `SubsetContribution` and secret-key/share imports reject blobs that previously deserialized.
- `dilithium`: `packing::unpack_sk` returns a `bool` canonicality flag; `poly::shiftl` is crate-private.

## Test plan

- [x] Each finding reproduced red before the fix and green after (error-variant or allocator/stack-probe assertions pin the mechanism, not just the outcome)
- [x] `cargo test --workspace` green (dilithium, hdwallet, threshold; unit + integration + e2e suites)
- [x] New regression suites: `threshold/tests/heap_zeroization.rs` (scanning allocator, 6 scenarios), `dilithium/tests/import_stack_zeroization.rs` (painted-stack probe, release builds), `hdwallet/tests/wormhole_zeroization.rs`
- [x] Release-mode runs for the stack-zeroization suite

---

> [!NOTE]
> **High Risk**
> Changes touch authentication-adjacent key import, DKG/signing/resharing state machines, and secret zeroization; behavioral tightenings may break callers relying on previously accepted malformed keys or private struct fields.
> 
> **Overview**
> Fourth security-review pass hardens **secret handling**, **import/deserialize boundaries**, and **threshold protocol state** across `dilithium`, `hdwallet`, and `threshold`.
> 
> **ML-DSA / dilithium:** `KeccakState` is now rate-typed (`Shake128State` / `Shake256State`) with absorb/squeeze phase tracking and canonical squeeze chunking. Secret-key import/serialize uses `Zeroizing` locals so `Copy` stack arrays are not left behind; `unpack_sk` / `eta_unpack` return canonicality and reject out-of-range η coefficients at import; `shiftl` is crate-private.
> 
> **hdwallet:** BIP32 `derive` bounds seed length and validates paths before HMAC; wormhole derivation wipes felt-encoded preimages before heap free.
> 
> **threshold:** Widespread `Zeroizing` on buffers and transport frames (DKG K_S, resharing Round 4, hyperball sampling); `DerivedKeyId` includes the derived public key TR hash; aborted resharing can recover the old share via `take_existing_share`; signing resets the session if Round 3 fails after commitments are aggregated; stricter Borsh/constructor checks on `PrivateKeyShare`, `SubsetContribution`, and `DkgConfig`; DoS guards (reveal length before hash, bounded `active_set`, etc.). Regression tests add heap/stack scanning allocators where relevant.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 243bb0d9c256a5ce915559bc6c16a20e28f85b1c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,23 @@ jobs:
         run: cargo build --workspace --locked
       - name: Test workspace
         run: cargo test --workspace --locked
+      # The stack-zeroization guard is gated to optimized builds
+      # (#![cfg(not(debug_assertions))]): debug codegen materializes compiler
+      # temporaries of the secret key that no source-level fix can wipe, so the
+      # debug `cargo test` above compiles this file to zero tests. The property
+      # it guards is entirely a function of codegen, making it the test most
+      # likely to silently regress on a toolchain bump — run it in release, and
+      # fail if zero tests ran so a gating/rename change can't turn the guard
+      # back into documentation.
+      - name: Optimized-build regression tests (stack zeroization)
+        run: |
+          set -o pipefail
+          cargo test --release --locked -p qp-rusty-crystals-dilithium \
+            --test import_stack_zeroization 2>&1 | tee stack-zeroization.log
+          grep -Eq "test result: ok\. [1-9][0-9]* passed" stack-zeroization.log || {
+            echo "::error::stack-zeroization tests were filtered out (0 tests ran)"
+            exit 1
+          }
 
   no-std-build:
     name: 📦 no_std Build (wasm32v1-none)
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -11,7 +11,9 @@ This PR addresses the fourth round of security review findings across the `dilit
 - DKG K_S relocation out of Copy-typed containers: `transition_to_round2` freed the `received_shared_secrets` map nodes with the secrets still in them (the state zeroizer runs too late — the field is already `None`), `pending_privates.pop()` left queued `Round1Private` bytes beyond the `Vec`'s shrunken length, and the buffered-privates drain consumed its map the same way (`4e6cb6d`);
 - the wormhole Poseidon preimage (`preimage_felts`) was `clear()`ed but never wiped; covered by a dedicated allocator-instrumented test in `hdwallet/tests/wormhole_zeroization.rs` (`7d63cf9`).
 
-**Stack copies of the ML-DSA-87 secret key are wiped in import/serialize paths.** `Keypair::from_bytes`, `SecretKey::from_bytes`, and `Keypair::to_bytes` left plaintext `[u8; SECRETKEYBYTES]` copies in dead stack frames because `[u8; N]` is `Copy`. All three now stage the secret in `Zeroizing` locals. A painted-stack probe test (`dilithium/tests/import_stack_zeroization.rs`, release-mode only — debug builds create compiler temporaries no source fix can wipe) verifies no copy survives (`2197133`).
+**Stack copies of the ML-DSA-87 secret key are wiped in import/serialize paths.** `Keypair::from_bytes`, `SecretKey::from_bytes`, and `Keypair::to_bytes` left plaintext `[u8; SECRETKEYBYTES]` copies in dead stack frames because `[u8; N]` is `Copy`. All three now stage the secret in `Zeroizing` locals, and the secret-bearing `to_bytes` methods return `Zeroizing<[u8; N]>` so the caller's copy self-wipes too (semver-breaking; a caller who simply drops the result leaks nothing). A painted-stack probe test (`dilithium/tests/import_stack_zeroization.rs`, release-mode only — debug builds create compiler temporaries no source fix can wipe) verifies no copy survives, and CI runs it in release mode with a zero-tests-ran guard.
+
+Known residual: the DKG private-send path (`pop_pending_private` → `poke` → `serialize_round1_private`) moves a `Round1Private` by value, and intermediate move temporaries holding K_S in dead stack frames are compiler-managed — the same class the dilithium probe pins. Heap copies on that path are covered; the stack side is noted in the code as follow-up.
 
 **Debug output no longer leaks key material.** The resharing `Action::SendPrivate` payload (serialized sub-shares) is redacted from the manual `Debug` impl (`df8dee3`).
 
@@ -32,7 +34,7 @@ This PR addresses the fourth round of security review findings across the `dilit
 ## Protocol state-machine correctness
 
 - Aborted resharing sessions no longer strand the old share: recovery restores it so the party can rejoin a retried session (`5e5c7ef`).
-- `round3_respond` failures **after** the peers' reveals were folded into the commitment aggregate (e.g. `recover_share` rejecting an active party outside the DKG set — only detectable at that point) now reset the session instead of leaving a poisoned aggregate in the "cleanly retryable" `AfterRound2` state, where a retry would double-count every peer's commitment. Pre-commit-point failures still leave the session clean for corrected retries (`243bb0d`).
+- `round3_respond` failures **after** the peers' reveals were folded into the commitment aggregate now reset the session instead of leaving a poisoned aggregate in the "cleanly retryable" `AfterRound2` state, where a retry would double-count every peer's commitment. Pre-commit-point failures still leave the session clean for corrected retries (`243bb0d`). The trigger for the reset — a Round 1 broadcast from a party outside the DKG participant set, previously only detected inside `recover_share` — is now rejected by `round2_reveal` before the commit point, so the failure is early and recoverable and the round-3 reset is defense in depth.
 - `DerivedKeyId` is bound to the DKG output public key, preventing cross-key derivation-ID collisions (`5599c69`).
 
 ## API contract enforcement
@@ -46,8 +48,8 @@ This PR addresses the fourth round of security review findings across the `dilit
 Behavioral tightenings downstream consumers may notice:
 
 - `hdwallet`: seeds outside 16..=64 bytes are now rejected by `ExtendedPrivKey::derive` (new `Error::InvalidSeedLength` variant). The Quantus SDK's `derive_hd_path` passes 64-byte BIP39 seeds and is unaffected.
-- `threshold`: resharing `Action::SendPrivate` now carries `Zeroizing<Vec<u8>>`; `DkgConfig` fields are private (use the accessors); `SubsetContribution` and secret-key/share imports reject blobs that previously deserialized.
-- `dilithium`: `packing::unpack_sk` returns a `bool` canonicality flag; `poly::shiftl` is crate-private.
+- `threshold`: resharing `Action::SendPrivate` now carries `Zeroizing<Vec<u8>>`; `DkgConfig` fields are private (use the accessors); `SubsetContribution` and secret-key/share imports reject blobs that previously deserialized; `ThresholdSigner::round2_reveal` rejects Round 1 broadcasts from parties outside the DKG set (previously deferred to Round 3); `ResharingProtocol::take_existing_share` is renamed `abort_and_take_existing_share` (the name now carries its abort side effect); `all_broadcasts_received` is no longer exported.
+- `dilithium`: `packing::unpack_sk` returns a `bool` canonicality flag; `poly::shiftl` is crate-private; `Keypair::to_bytes` and `SecretKey::to_bytes` return `Zeroizing` buffers (deref or `as_slice()` for access).
 
 ## Test plan
 
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -1,0 +1,57 @@
+# Security review round 4: secret hygiene, bounded deserialization, and protocol atomicity
+
+This PR addresses the fourth round of security review findings across the `dilithium`, `hdwallet`, and `threshold` crates. Every fix follows the same discipline: the finding was first reproduced with a failing regression test, then fixed, then re-verified (test red on the old code, green with the fix). 19 commits, 28 files, ~2600 insertions.
+
+## Secret hygiene (zeroization)
+
+**Heap intermediates are wiped before their memory is freed.** A new `threshold/tests/heap_zeroization.rs` suite installs a scanning global allocator that inspects every freed block for the scenario's secret pattern at `dealloc` time. It caught, and now guards, six scenarios:
+
+- `derive_dkg_contribution`'s share-digest linearization buffer, incoming DKG Round 1 private frames carrying K_S, and the hyperball sampling scratch buffer (`7d63cf9`);
+- the resharing protocol's `party_key` derivation buffer, Round 4 transport frames (both directions — `Action::SendPrivate` now carries `Zeroizing<Vec<u8>>`), sub-share derivation output, and per-polynomial hashing scratch buffers (`ce8175f`);
+- DKG K_S relocation out of Copy-typed containers: `transition_to_round2` freed the `received_shared_secrets` map nodes with the secrets still in them (the state zeroizer runs too late — the field is already `None`), `pending_privates.pop()` left queued `Round1Private` bytes beyond the `Vec`'s shrunken length, and the buffered-privates drain consumed its map the same way (`4e6cb6d`);
+- the wormhole Poseidon preimage (`preimage_felts`) was `clear()`ed but never wiped; covered by a dedicated allocator-instrumented test in `hdwallet/tests/wormhole_zeroization.rs` (`7d63cf9`).
+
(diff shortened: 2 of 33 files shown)


n13:
## Review

Round 4 of the security-review series: 20 commits, 28 files, +2621/−294 across `dilithium`, `hdwallet`, and `threshold`. Four themes — secret hygiene (heap + stack zeroization), bounded/validating deserialization, DoS hardening (no work before validation), and protocol state-machine atomicity.

The commit discipline is excellent: every fix is one commit, the message states the mechanism and the attacker path, and each carries a regression test that pins the *mechanism* (error variant, allocator hook, stack scan) rather than just the outcome. Several of the fixes are subtle and correct in ways that are easy to get wrong — the `Copy`-array leaks in particular (`[u8; N]` surviving a "move", `Vec::pop` leaving bytes past the shrunken length, `into_iter()` skipping element `Drop`s) are exactly the class of bug that reading the code casually would miss.

### Verification performed

On `illuzen/v12-round-4` @ `ce9c753`, Rust 1.93.0:

| Check | Result |
|---|---|
| `cargo check --workspace --all-targets` | clean |
| `cargo clippy --workspace --all-targets` | clean, no warnings |
| `cargo fmt --all --check` | clean |
| `cargo test --workspace` | green (incl. 3 `compile_fail` doctests: `fips202`, `poly`, `DkgConfig`) |
| `--test heap_zeroization` | 1 passed (all 6 scenarios) |
| `--test wormhole_zeroization` | 1 passed |
| `--test import_stack_zeroization` | **0 tests** in debug; 1 passed under `--release` |

Three things I specifically checked because they'd be silent, high-impact breaks:

**1. The `fips202` rework does not change any XOF output.** Making `keccak_squeezeblocks` continue the canonical stream mid-block is a behavioural change in principle. I checked all seven in-tree `shake{128,256}_squeezeblocks` call sites (`dilithium/src/poly.rs:340,348,583,605,628,642,1405`): every one operates on a state left at `pos == R` by `shake*_stream_init` / `shake*_finalize` / `absorb_once`, or by a preceding block-aligned `squeezeblocks`. The new `pos < R` branch is therefore never taken by existing code, and the `shake256()` simplification (whole-block loop → single `keccak_squeeze`) yields the identical stream. **No derived key, nonce, seed, or signature changes.** Worth stating explicitly in the PR description — this is the kind of change a reader will (rightly) be nervous about.

**2. `1u32 << total_parties` (`threshold/src/keys.rs:216`) cannot overflow.** `ThresholdConfig::new` runs first (line 195-ish) and `MAX_PARTIES = 6` (`threshold/src/error.rs:319`), so the shift is ≤ `1 << 6`. The comment already says this; I confirmed the ordering holds in the final code, since a reorder here would turn a validation check into a debug panic / silent mask-domain of 1.

**3. `max_frame_size()` really is dominated by the Round 2 broadcast.** Round 3 is `k × 4480` vs Round 2's `k × 5888`; Round 1 and Round 4 are fixed, and Round 4 (1 + 32 + 4627 = 4660 bytes) still fits under the smallest supported budget (k = 4 for (2,2) → 23,597). No legitimate frame is rejected by the new per-config cap.

---

### Findings

#### 1. The stack-zeroization guard never runs in CI (highest-value fix here)

`dilithium/tests/import_stack_zeroization.rs:30` is `#![cfg(not(debug_assertions))]`, and CI only runs `cargo test --workspace --locked` (`.github/workflows/ci.yml:73`) — i.e. debug. Confirmed empirically:

```
$ cargo test -p qp-rusty-crystals-dilithium --test import_stack_zeroization
running 0 tests
test result: ok. 0 passed; ...
```

The release run is green (I checked), but nothing keeps it that way. This test guards a property that is *entirely* a function of codegen, so it is precisely the test most likely to silently regress on a toolchain bump — and right now that regression is invisible. Suggest a step in `ci.yml`:

```yaml
- name: Optimized-build regression tests (stack zeroization)
  run: cargo test --release --workspace --locked
```

or, if a full release run is too slow, at minimum `cargo test --release -p qp-rusty-crystals-dilithium --test import_stack_zeroization`. Without this the test is documentation, not a guard.

#### 2. `from_bytes` still leans on optimizer elision; it doesn't have to

`Keypair::from_bytes` (`ml_dsa_87.rs:127`) and `SecretKey::from_bytes` (`:225`) end with:

```rust
Ok(SecretKey { bytes: *secret_bytes })   // *deref out of Zeroizing
```

The `*` is a copy out of the `Zeroizing` local into the struct field — the same "temporary that no source-level fix can wipe" the test's `#![cfg(not(debug_assertions))]` gate exists to tolerate. It happens to be elided in release, which is why the test passes, but the fix is one step short of unconditional. Since `SecretKey: ZeroizeOnDrop` and both functions live in the same module as the private `bytes` field, you can fill the destination in place and drop the reliance on codegen entirely:

```rust
let mut sk = SecretKey { bytes: [0u8; SECRETKEYBYTES] };
sk.bytes.copy_from_slice(bytes);
crate::sign::public_key_from_secret(&sk.bytes).ok_or(BadSecretKey)?;  // error path: sk drops, wiped
Ok(sk)
```

No `Zeroizing` needed, no intermediate array, and the error returns are covered by `SecretKey`'s own `Drop`. As a bonus this would likely let the test drop the debug-build gate, which also resolves finding #1.

#### 3. `to_bytes` still hands the caller an unwipeable plaintext key

`Keypair::to_bytes` now assembles into a `Zeroizing` buffer and returns `*result` (`:71`) — the library side is clean, but the *return value* is a bare `[u8; KEYPAIRBYTES]` containing the full secret key, and the caller has no `Drop` to lean on. The regression test has to wipe it by hand (`serialized.zeroize()`), which is a fair signal that the API shape is the actual hazard. Since this PR is already semver-breaking for `dilithium`, this is the cheap moment to return `Zeroizing<[u8; KEYPAIRBYTES]>` (and likewise `SecretKey::to_bytes`) and make the whole leak class unrepresentable rather than patched at each site. Your call whether that belongs in round 5.

#### 4. `eta_unpack`'s "branch-free" comment is not accurate

`dilithium/src/poly.rs:708-712`:

```rust
// Canonicality sweep after the fact: branch-free over the whole
// polynomial, so honest keys ... take a data-independent path
r.coeffs.iter().all(|&c| (-eta..=eta).contains(&c))
```

`Iterator::all` short-circuits on the first `false`. The *security* claim survives (honest keys always traverse fully, and the early-exit position is a function of attacker-supplied bytes, not of secret material) — but the comment asserts a property the code doesn't have, and comments in this crate are clearly treated as load-bearing. Either fix the comment or make it true:

```rust
r.coeffs.iter().fold(true, |ok, &c| ok & (-eta..=eta).contains(&c))
```

The aggregation in `unpack_sk` is correctly non-short-circuiting (`canonical &= …`), so this is the only gap.

#### 5. `all_broadcasts_received(&[], …) == true` — vacuous truth in a public quorum predicate

`threshold/src/keygen/dkg/state.rs:463` replaces the `len() - 1` underflow with a filter, and documents the empty list as "vacuously complete". For a predicate that is publicly re-exported (`keygen/dkg/mod.rs:206`) and takes a raw `&[ParticipantId]`, I'd argue the old release-build behaviour (never complete → stall) was the safer failure mode: "complete" lets a caller advance a round having received zero broadcasts. The `DkgConfig` hardening in this same PR means the in-crate path can no longer produce an empty list — which to me argues for making the function `pub(crate)` rather than for hardening it against inputs it can no longer receive. Failing closed (`if all_participants.is_empty() { return false }`) would be the alternative. Not a live vulnerability either way; flagging because the chosen semantics are the more permissive of the two.

#### 6. `SUBSET_BYTES` hardcodes `256` in two of three places

- `threshold/src/derivation.rs:137` — `(L + K) * 256 * size_of::<i32>()`
- `threshold/src/keygen/dkg/protocol.rs:1951` — same
- `threshold/src/resharing/protocol.rs:2434` — `(L + K) * N as usize * …`

`N == 256` today, so all three are correct. But the whole point of these constants is that an under-sized capacity silently reintroduces the growth-realloc leak they exist to prevent — a parameter change would break two of them quietly and only the allocator test would catch it. Use `N as usize` in all three.

#### 7. Round-3 reset closes a real hole but makes session-kill cheap; the root cause is upstream

The `round3_respond` change is right: leaving a mutated `w_aggregated` in a "cleanly retryable" `AfterRound2` was worse than resetting, and the reset matches the existing `aggregate_reveals` failure path exactly. The test that reaches it through the *documented* deferred check (rebranding party 1's valid commitment as ghost party 99) is a nice piece of work.

Note the residual though: the trigger is that `round2_reveal` accepts a Round 1 broadcast from a party outside `dkg_participants` — the check is deferred by design — so any peer whose broadcast the caller forwards can now force a full session restart instead of a recoverable error. That's a strictly better trade than a poisoned aggregate, but the clean fix is to validate the claimed party IDs against `dkg_participants` in `round2_reveal`, *before* the commit point, so the failure is early and the session stays retryable. Worth a round-5 item rather than a change here.

#### 8. These are major-version bumps, and nothing enforces that

Breaking surface, for the record: `DkgConfig` fields → private; `DkgAction::SendPrivate` / resharing `Action::SendPrivate` payload → `Zeroizing<Vec<u8>>`; `DerivedKeyId::new` signature + new public field; `packing::unpack_sk` → `bool`; `poly::shiftl` / `polyvec::k_shiftl` → `pub(crate)`; `KeccakState` → const-generic; new `hdwallet::Error::InvalidSeedLength` variant; behavioural rejection of previously-accepted keys, shares, `SubsetContribution` blobs, and seeds outside 16..=64.

The release workflows take `version_type` as a manual `workflow_dispatch` input, so nothing derives this from the diff — whoever cuts the release has to *know* to pick `major` for all three crates. The "Compatibility notes" section is good; consider also (a) noting the required bump explicitly there, and (b) marking `hdwallet::Error` / `KeyParsingError` `#[non_exhaustive]` so future variants stop being breaking.

#### 9. Minor

- `signing_protocol.rs` header check: gating the SSID comparison on `data.len() <= self.max_frame_size()` doesn't change what's rejected — it only routes oversized wrong-session frames to `Err(MalformedMessage)` instead of the silent `Ok(())` ignore. That's a defensible choice, but the comment explains it as a necessity ("fall through to the deserializer") rather than as the deliberate error-vs-ignore decision it is.
- `threshold/src/resharing/README.md` gained the abort-recovery paragraph but the integration sketch still implies a plain `Vec<u8>` payload for `Action::SendPrivate`. Worth a line so integrators hitting the type change find the reason next to the code.
- `take_existing_share()` transitioning an in-flight session to `Failed` as a side effect is correct and documented, but surprising for a method that reads as an accessor at the call site. `abort_and_take_existing_share()` (or similar) would carry the semantics.
- Residual stack copies in the DKG private-send path: `pop_pending_private` clones a `Round1Private`, which then moves through `serialize_round1_private` into `DkgMessage`. The final binding is wiped (`Round1Private: ZeroizeOnDrop`), but intermediate move temporaries holding K_S are not. Same class as the `dilithium` fix in this PR, out of scope for the heap work — noting it so it doesn't get lost.

---

### Verdict

**Approve.** The security content is sound, the tests genuinely pin the mechanisms rather than the symptoms, the workspace is green, and I verified independently that the `fips202` rework does not perturb any XOF output — the one change here that could have broken key/signature compatibility silently.

Nothing in findings 2–9 blocks merge. **Finding 1 I'd fix before merge** — it's a one-line CI addition, and without it the release-gated stack-zeroization test is inert in the only place it would ever catch a regression. Findings 2–4 and 6 are small, mechanical, and would tighten fixes this PR already makes; 5, 7, 8 are judgement calls worth a reply rather than a change.

<sub>Reviewed with Claude Code — branch checked out and built locally; all commands above run against `ce9c753`.</sub>

n13 (approved):
LGTM

PR #105, opened 2026-07-26
Type-enforced invariants, API contracts, and protocol liveness

illuzen:
# Security review round 5: type-enforced invariants, API contracts, and protocol liveness

This PR addresses the fifth round of security review findings across the `dilithium` and `threshold` crates. As in round 4, every fix was first reproduced with a failing test (a runtime regression test or a `compile_fail` doctest, as appropriate), then fixed, then re-verified red-on-old / green-on-new. 9 commits, 20 files, ~590 insertions.

The unifying theme this round is moving invariants out of documentation and runtime checks and into the type system: private fields with validating constructors, fixed-size array parameters instead of slice-plus-assert, removed trait impls, and `pub(crate)` scoping — so misuse fails at compile time instead of silently corrupting state at run time.

## Type-enforced invariants (dilithium)

**`Keypair` can no longer hold mismatched halves** (`f866801`). The secret/public correspondence was checked only in `Keypair::from_bytes` while both fields were `pub`, so a caller could assemble or mutate a keypair that signs under one key and advertises another. The fields are now private with `secret()`/`public()` accessors, and the only ways to build a keypair — `generate`, `from_bytes`, and the new `from_parts` — all establish the invariant (`from_parts` re-derives the public key from the secret and rejects a mismatch). `Keypair` also gained a `Zeroize` impl that wipes its secret half. Semver-breaking; all in-workspace call sites migrated. `compile_fail` doctests pin the field privacy.

**`KeccakState` is no longer `Clone`** (`9bd95f2`). The sponge state's own docs say copies must be prevented — signing absorbs the private key seed `K` into this state to derive the mask seed ρ' — yet the type derived `Clone`, so a full duplicate of secret-absorbed state was one call away, invisible to the original's `ZeroizeOnDrop`. Nothing in the workspace clones a state (reuse goes through `init()`/`default()`), so the derive is removed and pinned with a `compile_fail` doctest.

**`unpack_sig` is `#[must_use]`** (`32fb884`), matching its sibling `unpack_sk`. The boolean is the only signal that the hint encoding is canonical, and all output buffers are written even on `false`, so a caller that drops the flag proceeds with a structurally invalid signature. Pinned with a `compile_fail` doctest under `deny(unused_must_use)`.

**Buffer sizes and coefficient bounds are part of the signature** (`4ed5dc2`).

- `poly::pointwise_montgomery` now documents and `debug_assert!`s `montgomery_reduce`'s input domain (`|a[i]·b[i]| ≤ 2^31·Q`); outside it the reduction returns a silently wrong residue, not a panic. NTT-domain inputs satisfy it comfortably.
- `poly_pack_w` (threshold) gained the debug panic its docs promised — tightened to the true contract `[0, Q)`, since a negative coefficient wraps via `as u32` into 23 bits of garbage — and its buffer is now `&mut [u8; POLY_Q_SIZE]`, deleting the release-mode `assert!`.
- `pack_w_dilithium` takes `&mut [u8; SINGLE_COMMITMENT_SIZE]`; both packing loops carve exact chunks with `as_chunks_mut`, removing the defensive offset arithmetic. `SINGLE_COMMITMENT_SIZE` has a single authoritative `K * POLY_Q_SIZE` definition replacing the hand-written `8 * 736`.
- `poly::challenge`'s seed is typed `&[u8; C_DASH_BYTES]`: a wrong-length seed would not fail — it would silently derive an off-domain challenge.

## Correctness (dilithium)

**`poly::decompose` output order matched neither its docs nor its sibling** (`5e9c341`). It destructured rounding's `(low, high)` tuple backwards, so the parameter named `a1` received the *low* bits — and `k_decompose` silently compensated with an extra `swap()` that `k_power2round` doesn't need. In-crate behavior was correct, but an external integrator calling `decompose(&mut w1, &mut w0)` by analogy with the documented `power2round` convention got swapped halves and silently invalid signatures. The destructuring is flipped and the compensating swap deleted; the NIST KAT suite passing byte-exact proves the two changes net out to identical in-crate behavior.

## Wire-format hardening (threshold)

**`Round4Broadcast.transcript_signature` is bounded at the type's own deserializer** (`1914037`). The manual Borsh decoder capped the `partial_public_keys` map at `MAX_SUBSETS` but read the signature as an unconstrained `Vec<u8>`; the 256 KiB frame gate lives only in the private `Dkg::message` path, so integrators deserializing the exported wire type directly had no protection against a fully-supplied oversized payload. A new `MAX_TRANSCRIPT_SIGNATURE_SIZE` (64 KiB — an order of magnitude above ML-DSA-87's 4627-byte signatures, well under the frame gate) is enforced in `deserialize_reader`, which also reads incrementally so a truncated payload cannot force an up-front allocation of the claimed length.

## Protocol liveness (threshold)

**Resharing no longer strands a session on a delayed non-dealer Round 3 broadcast** (`f318244`). New-only and non-active parties skip `Round3Waiting`'s `have_enough_round3` gate, so they could advance to Round 5 on dealer-only Round 4 traffic while an active non-dealer's Round 3 broadcast was still in flight. The Round 3 buffering window closes at that transition, so the late arrival was dropped forever and the session died in Combining ("Missing Round 3 broadcast from active party"), where the transcript needs Round 3 from *every* active member. Since active non-dealers exist in every config with `t_old < n_old`, a transport adversary could deny every handoff by delaying one message per session. `handle_round4_waiting` now holds until Round 3 broadcasts from all active members are present. A regression test drives a full 2-of-3 handoff with the harness withholding the non-dealer's Round 3 from a new member until the network goes quiescent, then delivering it and asserting completion.

## API scoping (threshold)

**`all_private_messages_received` is `pub(crate)`** (`cfc6845`), mirroring `all_broadcasts_received`. Both predicates return "complete" for empty input (vacuous truth), which is sound only for `DkgConfig`-validated inputs; exporting one while restricting the other for exactly that reason was contradictory. Both are now pinned by `compile_fail` doctests in `DkgConfig`'s invariants section.

## Documentation

**The unvalidatable nonce seed `K`** (`b6244f9`). Import validation binds `rho`/`s1`/`s2`/`t0`/`tr`, but the packed `K` field *cannot* be validated: it is independent entropy with no stored commitment, and under FIPS 204 every 32-byte value forms a valid key. An attacker with write access to stored key blobs can therefore silently replace `K`, converting known-`K` deterministic signatures into known-mask signatures from which the secret vector is recoverable. Any in-format fix would break FIPS 204 compliance or KAT compatibility, so the boundary is documented instead: callers must integrity-protect secret-key storage or use hedged signing, which keeps ρ' unpredictable even for a known `K`.

## Verification

- Full workspace test suite green, including the long-running signing and resharing integration suites and the release-mode stack-zeroization probe.
- NIST KAT tests pass byte-exact (pins the `decompose`/`k_decompose` refactor and confirms the debug asserts sit outside release codegen).
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean.
- Semver: breaking for `dilithium` (`Keypair` field privacy, `KeccakState: !Clone`, `challenge` seed type) and `threshold` (`Round4Broadcast` deserialization bound, removed re-export). All changes ride the already-breaking release.

---

> [!NOTE]
> **Medium Risk**
> Breaking public API on `Keypair` and wire deserialization, plus signing-adjacent polynomial and protocol state changes; mitigated by KATs, regression tests, and mostly defensive hardening rather than new crypto logic.
> 
> **Overview**
> Moves **dilithium** invariants into the type system: **`Keypair`** halves are private with **`secret()`/`public()`**, **`from_parts`**, and **`Zeroize`**; **`KeccakState`** drops **`Clone`**; **`unpack_sig`** is **`#[must_use]`**; **`poly::challenge`** takes a fixed-size seed; packing paths use fixed buffer types and **`debug_assert!`** on **`pointwise_montgomery`** / **`poly_pack_w`**.
> 
> Fixes **`poly::decompose`** so **`a1`/`a0`** match **`power2round`** and removes the compensating **`swap`** in **`k_decompose`**. Documents the unvalidatable nonce seed **`K`** and hedged vs deterministic signing.
> 
> **Threshold**: **`Round4Broadcast`** deserialization caps **`transcript_signature`** at **`MAX_TRANSCRIPT_SIGNATURE_SIZE`**; **`all_private_messages_received`** is **`pub(crate)`** like the broadcast quorum helper; commitment packing uses shared **`POLY_Q_SIZE`** / **`SINGLE_COMMITMENT_SIZE`** chunks.
> 
> **Resharing** now blocks leaving **`Round4Waiting`** until every active member’s Round 3 broadcast is present (fixes sessions that advanced on dealer-only traffic and then dropped late non-dealer Round 3). Workspace call sites migrate to the new **`Keypair`** accessors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cfc68453bf18575a01ead70577e5bf416431f694. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13 (approved):
# Review

## Overview

Round 5 of the security-review series: 9 commits across 20 files that move invariants from docs/runtime checks into the type system (`Keypair` field privacy + `from_parts`, `KeccakState: !Clone`, `#[must_use]` on `unpack_sig`, fixed-size buffer/seed types), fix the `poly::decompose` output-order inversion, bound `Round4Broadcast.transcript_signature` at the wire-type boundary, and close a real resharing liveness hole where new-only/non-active parties could exit `Round4Waiting` before all active members' Round 3 broadcasts had arrived.

I checked out the branch and verified the claims independently rather than taking the PR description at face value. Everything reproduced.

## What I verified

**`decompose`/`k_decompose` refactor is behavior-neutral.** `rounding::decompose` returns `(a0, a1)` = (low, high), so the flipped destructuring puts the high part in `a1` as documented, and deleting the compensating `swap` in `k_decompose` nets out to identical in-crate behavior. Confirmed empirically: the NIST KAT/ACVP integration suite passes byte-exact on the branch. The fix genuinely removes an external-integrator footgun — `sign.rs:445` is the only in-crate caller, but any external caller following the documented `power2round` convention previously got swapped halves.

**The resharing gate is deadlock-safe and the regression test is honest.** Traced the state machine: `Round4Waiting` is only reachable via `Round3Generate`, which requires `have_act_entropy_data()` — false until `active_set` is `Some` — so `have_enough_round3()` can never permanently return `false` on an unknown active set. Round 3 broadcasts are accepted from protocol start through `Round4Waiting` and every active member broadcasts to all parties, so the gate condition is eventually satisfiable for every party it binds. I also ran the red-on-old check: with the gate removed, `test_new_member_waits_for_nondealer_round3` fails with exactly the reported `Missing Round 3 broadcast from active party 2 for transcript`; with the gate, the full 69-test resharing suite passes, so the gate doesn't stall any existing scenario. Gate placement is right — dealers drain their pending Round 4 sends *before* the gate, so outbound traffic is never blocked.

**The Borsh bound matches the wire format.** Borsh encodes `Vec<u8>` as a u32 LE length prefix + raw bytes, so the manual `u32` read + `read_length_prefixed` is wire-compatible with the derived serializer (the at-bound round-trip test confirms it). `read_length_prefixed` caps per-step allocation at 64 KiB, so a truncated frame can't force an up-front allocation of the claimed length.

**API/tooling.** Toolchain is pinned at 1.93.0, so the new `as_chunks_mut` (1.88) and `first_chunk` (1.77) usages are fine. `sign.rs:627` (verify) checks `unpack_sig`'s flag, so `#[must_use]` breaks no in-crate caller. `SecretKey` is `ZeroizeOnDrop`, so the new `Keypair::zeroize` doc claim about drop-time wiping holds. All workspace call sites of the now-private `Keypair` fields are migrated.

**Test results on the branch** (macOS, rustc 1.93.0):

- dilithium: 112 lib tests, 7 doctests (incl. the new `compile_fail` pins), keygen-zeroization, and the 10-test NIST KAT/ACVP suite — all pass
- threshold: 257 lib tests, 13 doctests, 67 resharing tests, 48 signing integration tests, 15 buffer-bounds tests, signer-state test — all pass
- hdwallet: 45 tests pass
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: clean

## Minor findings (non-blocking)

1. **Stale comment left behind by this PR** — `dilithium/src/sign.rs:215`: "Note signature depends on std because k_decompose depends on swap which depends on std". This PR removes that `swap`, so the comment is now dangling (it was also inaccurate before — `core::mem::swap` never required std). Worth deleting here since this is the PR that removed the swap.
2. **Silent truncation in the repacked commitment loops** — `generate_round1`/`pack_round1_commitment` now `zip` chunks with `w_commitments`; if `w_commitments.len() < k_iterations` the trailing chunks stay zeroed silently. This preserves the old guarded behavior exactly (not a regression), but a `debug_assert_eq!(w_commitments.len(), k_iterations)` would surface the mismatch the old defensive bounds check was quietly tolerating.
3. `MAX_TRANSCRIPT_SIGNATURE_SIZE` (64 KiB) happens to equal `read_length_prefixed`'s internal `CHUNK`, so the "incremental" read is a single `read_exact` today. Fine as-is; just noting the headroom claim only becomes incremental if the bound ever grows.

## Verdict

**Approve.** Every substantive claim in the PR description reproduced under independent verification: the decompose inversion was real and the fix is KAT-pinned byte-exact; the resharing liveness hole was real (red-on-old confirmed) and the gate is provably deadlock-safe; the wire bound is format-compatible and tested at the boundary. The semver breaks are intentional, documented, and ride an already-breaking release. The two nits above are cleanup-level and can land here or in a follow-up.

PR #106, opened 2026-07-27
All ML-DSA Variants

illuzen:
# Multi-variant ML-DSA support (44 / 65 / 87) across dilithium, threshold, and hdwallet

## Summary

This PR extends the workspace from ML-DSA-87-only to all three FIPS 204 parameter
sets, selected at compile time via additive cargo features. It lands in three parts:

1. **`qp-rusty-crystals-dilithium`** — the core implementation is genericized
   over the parameter set (const generics for K/L, variant-parameterized
   primitives), with per-variant frontends `ml_dsa_44` / `ml_dsa_65` /
   `ml_dsa_87` behind matching cargo features, validated against ACVP and
   PQCrystals NIST KATs for every variant.
2. **`qp-rusty-crystals-threshold`** (bumped to **2.0.0**) — the threshold
   signing, DKG, derivation, and resharing protocols are driven by a new
   `params` module that re-exports the active variant's constants and ships
   per-variant Monte-Carlo calibration tables (hyperball radii, `k_iterations`,
   and resharing enlargement κ).
3. **`qp-rusty-crystals-hdwallet`** (bumped to **3.1.0**) — HD key derivation
   gains per-variant modules (`ml_dsa_44` / `ml_dsa_65` / `ml_dsa_87`) that can
   coexist in one build; the top-level API stays the ML-DSA-87 one.

Nothing in the threshold crate has been deployed, so **no wire-format
compatibility is preserved**: the SSID version is bumped to 3, the suite ID is
bound into the SSID, and one deprecated constant is removed.

Signatures remain byte-compatible with standard ML-DSA verification for the
active parameter set on all three variants (pinned by a direct cross-crate test).

The branch also merges current master (the `Keypair` API-hardening / typed-packing
PR), adds a CI matrix for feature-exclusive 44/65 builds, vendors golden HD
derivation vectors for 44/65, and recalibrates the 44/65 threshold tables at
high Monte-Carlo sample counts (see Part 5).

74 files, +8,589/−2,270. Each commit builds standalone (see "Commit guide" for
one caveat about test targets on intermediate commits).

---

## Feature selection

Dilithium and threshold expose additive features with priority **87 > 65 > 44**
when several are enabled (so a workspace `--all-features` build gets ML-DSA-87).
The hdwallet crate is the exception: its features are additive-coexisting — each
enabled variant gets its own module, since a wallet may hold keys at several
security levels at once.

| Feature | NIST category | Status |
|---|---|---|
| `ml-dsa-87` (default) | 5 (~256-bit) | Production-calibrated threshold tables (pre-existing, unchanged) |
| `ml-dsa-65` | 3 (~192-bit) | Calibrated threshold tables (2k-sample grid, 8k-sample K refine); reshares at κ = 1 everywhere |
| `ml-dsa-44` | 2 (~128-bit) | Calibrated threshold tables (same method); reshare into (4,6) fails closed |

Suite IDs (bound into the threshold SSID and the resharing SSID): `1` = ML-DSA-87,
`2` = ML-DSA-44, `3` = ML-DSA-65. `THRESHOLD_SSID_VERSION = 3`.

---

## Part 1 — dilithium crate

- **Const-generic core**: `Polyveck`/`Polyvecl` collapsed into a single
  `Polyvec<N>`; the k_*/l_* function shims are removed. Packing, the signing
  driver, and the variant-dependent primitives (`ETA`, `GAMMA1`, `GAMMA2`,
  `TAU`, rounding/decompose/hints) are parameterized over the variant's
  constants rather than hardcoded to 87.
- **Per-variant frontends**: `ml_dsa_44`, `ml_dsa_65`, `ml_dsa_87` modules
  (Keypair/PublicKey/SecretKey, sizes, `MAX_MESSAGE_SIZE`) behind their cargo
  features. The legacy top-level re-exports (`params::K`, etc.) keep their
  ML-DSA-87 values so existing call sites are unaffected.
- **Validation**: vendored ACVP keygen/siggen/sigver vectors for 44/65 with a
  generalized harness (87 coverage unchanged), plus PQCrystals NIST KATs for
  44/65. CI runs the ACVP suites with `--all-features`.
- **Benchmarks** for all three variants behind their features.

## Part 2 — threshold crate (2.0.0)

- **`threshold/src/params.rs`**: single source of truth for the active
  parameter set. Re-exports the active variant's constants (K, L, ETA, GAMMA1,
  GAMMA2, TAU, sizes…), defines `SUITE_ID`, `THRESHOLD_SSID_VERSION`, derived
  size bounds (`SINGLE_COMMITMENT_SIZE`, `MAX_COMMITMENT_DATA_SIZE`, …), and
  per-variant calibration tables:
  - `K_ITERATIONS` — parallel signing attempts per `(t, n)`
  - `HYPERBALL` — rejection-sampling radii `(r, r', ν)` per `(t, n)`
  - `RESHARING_KAPPA` — resharing enlargement κ per `(t, n)` (see Part 3)

  87's tables are inline (unchanged values); 44/65 tables live in
  `params_tables_44.rs` / `params_tables_65.rs`, generated by
  `scripts/compute_hyperball_params.py --variant {44,65}`.
- **`threshold/src/mldsa.rs`**: feature-selected frontend types
  (`MlDsaPublicKey`, `MAX_MESSAGE_SIZE`, test-only `Keypair`/`SecretKey`).
- **Protocol parameterization**: signing, config, keys, broadcasts, signer
  buffer sizes, DKG (dealer and distributed), derivation, and resharing all
  read from `params` instead of hardcoded 87 values. `decompose_coefficient`
  is γ₂-aware (16 high-bit values for γ₂ = (Q−1)/32, 44 for (Q−1)/88).
- **Domain separation**: SSID computation now absorbs
  `THRESHOLD_SSID_VERSION = 3` and the suite ID, so sessions on different
  parameter sets can never be confused. The resharing suite ID follows the
  active set; the deprecated `RESHARING_SUITE_ML_DSA_87` alias is removed.

## Part 3 — review hardening and resharing calibration

A commit-by-commit review of Parts 1–2 surfaced and fixed the following.

### Correctness fixes

- **Test targets didn't build on 44/65**: `resharing_tests.rs`, the round2
  input-validation DoS test, and the comparison benchmark had hardcoded
  K=8/L=7 fixtures or imported `ml_dsa_87` directly. All test/bench targets now
  compile and pass on all three variants. The crate publicly re-exports
  `MAX_MESSAGE_SIZE` for callers that need the bound.
- **Overshoot regression math was 87-only**: the resharing variance tests
  computed the Mithril §3.4 base bound `B` with hardcoded `k=8, ℓ=7, τ=60,
  Var=2`, silently invalidating the `overshoot ≤ κ` regression assertion on
  other variants. Now computed from the active params.
- **Decompose pinned exhaustively**: the multi-variant decompose rewrite
  changed the low-bits convention from "always `a₀ + q`" to "`a₀ mod q`"
  (every consumer normalizes mod q). A new exhaustive test over the full
  `[0, Q)` domain asserts the FIPS 204 Decompose contract for both γ₂
  variants **and** bit-for-bit (mod q) equivalence with the original audited
  ML-DSA-87 implementation, kept verbatim in the test as an oracle.
- **Standard-verifier compatibility pinned directly**: a new test verifies a
  threshold signature using `qp_rusty_crystals_dilithium`'s `PublicKey::verify`
  on raw bytes (not via `threshold::verify_signature`, which happens to
  delegate to it), including tampered-message and context-binding rejections,
  on all three variants.

### Resharing parameters (the main security fix)

The resharing Round-5 guard accepts recovered partials up to `κ·B`, where the
security argument requires the hyperball radii `(r, r')` to be enlarged by the
**same** κ (the radius condition `r'² = r² + B² + 2rB/φ` is scale-invariant, so
per-sample leakage ε is preserved). Before this PR's review pass,
`resharing_norm_enlargement()` returned the 87-measured κ (1.10/1.15/1.25) on
**every** variant while the 44/65 tables shipped base (κ = 1) radii — i.e. the
guard on 44/65 was **fail-open** relative to the proof.

Fix: κ moved into the per-variant params tables with the invariant that any
κ > 1 entry has its radii/K derived at the enlarged radius. Honest-reshare
overshoots were measured per variant (`test_recovered_partial_variance_*` built
with each feature; fixed point over all signing sets):

| Config | 87 (shipped κ) | 44 (new κ) | 65 (new κ) |
|--------|----------------|------------|------------|
| 2-of-2 | 0.780 (1.00) | 0.794 (1.00) | 0.620 (1.00) |
| 2-of-3 | 0.810 (1.00) | 0.827 (1.00) | 0.672 (1.00) |
| 2-of-4 | 0.961 (1.10) | 0.991 (1.10) | 0.869 (1.00) |
| 3-of-5 | 1.012 (1.15) | 1.023 (1.15) | 0.809 (1.00) |
| 4-of-6 | 1.163 (1.25) | 1.166 (**1.00, fail-closed**) | 0.917 (1.00) |

- **ML-DSA-87**: unchanged.
- **ML-DSA-44**: overshoots track 87 almost exactly, so (2,4)/(3,5) ship the
  same κ with radii scaled and `K` re-derived by Monte Carlo at the enlarged
  radii (final values after the Part-5 recalibration: (2,4) K = 14, (3,5)
  K = 488 — 44's tight verification ceilings, γ₁ = 2¹⁷ and γ₂ = (Q−1)/88, make
  enlargement far more expensive than 87's K = 60). **(4,6) cannot be
  enlarged**: at κ = 1.25 the MC estimate is K ≈ 10⁶, which is infeasible.
  It ships κ = 1 and base signing radii, so reshares *into* a 4-of-6 committee
  are rejected by the Round-5 guard — fail closed, never fail open. Freshly
  dealt/DKG 4-of-6 committees sign normally. A dedicated test pins the
  rejection.
- **ML-DSA-65**: overshoots are below 1 everywhere because its η = 4 keygen
  variance (20/3) dwarfs the η = 2-tuned split noise, so every committee
  reshares at κ = 1 with base tables. The flip side — documented at
  `SPLIT_NOISE_NUM_X256` and in the security proof — is that the aggregated
  split noise is ~3.3× below keygen variance, so the a-posteriori
  "reshared shares look like a fresh keygen sharing" hiding argument is
  quantitatively weaker on 65 than on the η = 2 sets. Re-tuning the intensity
  for η = 4 (which would raise overshoots and require κ > 1 radii
  re-derivation) is flagged as future calibration work.

### Documentation

- `resharing/SECURITY_PROOF.md` gains a **"Parameter-Set Scope"** section
  (structural argument is variant-independent; calibration is per variant;
  measured overshoot tables; reproduction commands).
- `resharing/README.md` and the crate README carry matching per-variant status
  notes; the hyperball script's resharing computation is generalized to any
  variant (`RESHARING_DATA`, `--resharing-only`).

---

## Part 4 — hdwallet crate (3.1.0)

The wallet's ML-DSA surface was one call: `Keypair::generate` on path-derived
entropy. That call is now available per variant.

- **Per-variant modules**: features `ml-dsa-44` / `ml-dsa-65` / `ml-dsa-87`
  (default: 87) each expose a module with `derive_key_from_seed` /
  `derive_key_from_mnemonic` returning that variant's `Keypair`. Unlike
  threshold, the modules **coexist** — no priority selection — so one build can
  derive keys at multiple security levels.
- **Shared core factored out**: path validation + HMAC-SHA512 tree derivation
  live in one internal `derive_entropy_from_seed`; the variant modules are thin
  macro-stamped wrappers over it. Mnemonic handling and the wormhole module are
  parameter-set independent and build with no ML-DSA feature at all.
- **Key separation is by construction**: FIPS 204 keygen absorbs `(k, ℓ)` into
  the seed expansion (`H(seed ‖ k ‖ ℓ)`), so the same path-derived entropy
  yields independent keys per variant. Deriving several variants from one path
  is sound; a test pins that the resulting keys diverge.
- **Backward compatible**: the top-level `derive_key_from_seed` /
  `derive_key_from_mnemonic` keep their signatures and remain byte-identical to
  `ml_dsa_87::*` (pinned by test). The historical test suite and its vendored
  vectors are 87-only and are gated on that feature; a new multi-variant suite
  covers determinism, sign/verify round-trips, and cross-variant independence.
- **Golden vectors**: vendored derivation vectors for 44/65 (`test_vectors_44.rs`
  / `test_vectors_65.rs`, regenerable via an `#[ignore]` test) pin secret/public
  key prefixes and lengths for fixed mnemonic+path inputs, so a silent change in
  the derivation-to-keygen pipeline fails CI.
- The `tests` integration crate now opts into all three hdwallet variants
  (the workspace pin uses `default-features = false`).

---

## Part 5 — post-review follow-ups

- **Merge with master**: the branch absorbs the API-hardening/typed-packing PR
  (#105): private `Keypair` fields with `secret()`/`public()`/`from_parts`,
  `#[must_use]` `unpack_sig`, fixed-size packing buffers, and the new
  `poly::challenge` signature — all re-reconciled with the multi-variant
  generics.
- **CI matrix**: the workflow now tests threshold and hdwallet under
  feature-exclusive `ml-dsa-44` and `ml-dsa-65` builds (previously only the
  dilithium crate was exercised per variant, so a 44/65-only regression in the
  higher crates could land green).
- **44/65 table recalibration**: the initial tables came from a 400-sample MC
  grid search, whose winner's-curse noise badly underestimated `K` for deep
  committees (44 (4,6): shipped K = 584 vs ~2,450 measured at 8,000 samples).
  The script's MakeHint/negacyclic-multiply hot loops were vectorized with
  NumPy (~150×; seeded-run equivalence with the scalar path verified), the full
  grid search re-run at **2,000 samples per point**, and `K` re-estimated at
  the chosen radii with **8,000 samples** and the usual 2× margin
  (`--refine-shipped`, which recovers κ-enlarged entries' acceptance exponent
  from the base radius). 87's audited tables are untouched.

## Notes for reviewers / auditors

- **Trust boundary of the numbers**: the 87 tables and κ are the pre-existing
  audited/production values, unchanged. The 44/65 signing tables are
  calibrated but **not audited** (2,000-sample MC grid for radii, 8,000-sample
  K refine, 2× margin). Regeneration is reproducible via
  `threshold/scripts/compute_hyperball_params.py`.
- **Invariant to check**: any `RESHARING_KAPPA` entry with κ > 1 must have its
  `HYPERBALL`/`K_ITERATIONS` entries derived at the enlarged radius. The
  variance regression tests assert measured overshoot ≤ κ from the same table
  the guard reads, so a κ lowered below the real overshoot (or a splitter
  regression) fails CI.
- **Decompose convention change** (low bits `a₀ mod q` vs the old `a₀ + q`) is
  intentional; both consumers (`w0_plus_diff` renormalization, hint
  computation) reduce mod q before use, and the exhaustive oracle test pins
  equivalence.
- **Deep committees are expensive on the smaller sets.** The honest 8k-sample
  estimates put 44 at (4,6) K = 2,450 / (5,6) K = 4,196 (~7–12 MB of commitment
  data per party per signature) and 65 at (4,6) K = 1,546 / (5,6) K = 2,444
  (~7–11 MB). This is intrinsic to their tighter γ₁/γ₂ ceilings, not a tuning
  artifact — if those committee shapes matter below category 5, budget
  accordingly.

## Testing

All run for each of `ml-dsa-44`, `ml-dsa-65`, `ml-dsa-87` (feature-exclusive
builds):

- `cargo test --lib`: green (255+ tests per variant).
- All integration suites (integration, coverage, derivation, error display,
  round2 DoS, resharing round5 DoS, resharing): green, re-run in full against
  the recalibrated tables. On 44 the (4,6) reshare suites are gated off by
  design and a dedicated fail-closed rejection test runs instead.
- `cargo clippy --all-targets -- -D warnings`: clean on all three variants.
- Dilithium ACVP + NIST KAT suites for all variants (run in CI with
  `--all-features`).
- Hyperball script: vectorized evaluator verified byte-identical to the scalar
  original on seeded runs for both 44 and 65.

hdwallet matrix (all green, clippy `-D warnings` clean):

- default (`std` + `ml-dsa-87`): full historical suite + variant suite.
- `std,ml-dsa-44` / `std,ml-dsa-65` exclusive: variant + shared + golden-vector
  suites.
- all three variants together: includes cross-variant independence tests.
- `std` only (no ML-DSA): wormhole/mnemonic surface builds and passes.

CI additions: the wasm32v1-none no_std job passes
`--features ml-dsa-44,ml-dsa-65,ml-dsa-87` explicitly for dilithium and
hdwallet (with the frontends feature-gated, a bare `--no-default-features`
build compiles only the generic core and would mask a no_std regression in the
public signing API); the test workflow gains feature-exclusive 44/65 jobs for
threshold and hdwallet.

## Commit guide

**dilithium genericization** (`815d6e5..12fa431`): const-generic `Polyvec`,
parameterized primitives/packing/signing, per-variant params modules and
frontends.

**dilithium validation** (`3fb57da..28278e1`): ACVP vendoring + harness,
benchmarks, CI, docs, alias removal, NIST KATs.

**threshold multi-variant** (`440c7f3..5261a48`): 2.0.0 + features, params
module + frontend, γ₂-aware decompose, signing/config/keys/DKG/derivation/
resharing parameterization, README + script.

**review hardening + calibration** (`485e871..f398b46`): 44/65 test-target
fixes, exhaustive decompose oracle test, per-variant κ (fail-open fix),
ML-DSA-44 resharing calibration + (4,6) fail-closed, scope docs, direct
standard-verifier compatibility test.

**hdwallet multi-variant** (`717fbab..0a5bbb8`): per-variant derivation
modules, README rewrite, no_std CI fix.

**post-review follow-ups** (`50c916c..4625f6a`): master merge (API hardening),
CI matrix for exclusive 44/65 builds, golden HD derivation vectors,
high-sample table recalibration.

Intermediate-commit caveat: `resharing_tests.rs` fails to compile under 44/65
between `440c7f3` and `c553cea` (inherited hardcoded fixtures, fixed within the
series); the default 87 build compiles all targets at every commit.


n13 (changes requested):
## Verdict

**Request changes** — but narrowly. The cryptographic core of this PR is in excellent shape: a line-by-line comparison of every genericized secret-handling primitive against the audited master baseline found **no constant-time regressions, no memory/zeroization regressions, and no parameter errors**, and the threshold invariants (κ enlargement, fail-closed paths, bounds) all check out. Two things should be fixed before merge: a domain-separation gap that contradicts the PR's own stated security model, and one cross-crate duplication of security-critical bit-twiddling.

Review method: four independent deep passes (dilithium constant-time/correctness vs master; dilithium memory/API hygiene; threshold protocol; hdwallet + whole-PR DRY sweep), plus independent re-runs of the test matrix: dilithium `--all-features` (ACVP KATs for all three variants), feature-exclusive 44/65 builds of threshold+hdwallet, workspace clippy `-D warnings`, hdwallet golden-vector suites, and the NIST KAT integration tests. All green locally, matching CI.

## Required

### 1. DKG SSID does not absorb `THRESHOLD_SSID_VERSION` / `SUITE_ID`

`compute_dkg_ssid` (`threshold/src/keygen/dkg/types.rs:723-757`) still hashes only `"THRESHOLD_DKG_SSID_V1" || t || n || participants || nonce` — unchanged from master. The signing SSID (`protocol/signing.rs:114-122`) and the resharing SSID both absorb version + suite; DKG does not, and neither does the derivation domain (`derivation.rs:68`, still `near-mpc-dilithium-dkg-contribution-v2` with no suite input). So two builds on different parameter sets with the same `(t, n, participants, session_nonce)` derive the **same** DKG SSID, and `h_commit`/transcript hashes inherit it — contradicting the description's claim that "sessions on different parameter sets can never be confused."

Exploitability is low: cross-variant DKG payloads fail shape-checked deserialization (`[[i32; 256]; L/K]` with L+K = 8/11/15), and derivation digests are over variant-shaped data. But a Round-1 commitment is an opaque 32-byte hash and *is* replayable across suites under an identical SSID. It's a one-line fix per site, nothing is deployed, and the wire format is already being broken in this PR — this is the moment.

### 2. `decompose_coefficient` duplicates dilithium's `rounding::decompose` verbatim across crates

`threshold/src/protocol/primitives.rs:217-235` re-copies the exact fixed-point magic of `dilithium/src/rounding.rs:49-68` — `(a1 * 1025 + (1 << 21)) >> 22`, `(a1 * 11275 + (1 << 23)) >> 24`, the branchless 44→0 wrap — including its own duplicated `GAMMA2_32`/`GAMMA2_88` consts, differing only in a 2-line wire-convention conversion at the end. Root cause: `dilithium/src/lib.rs:105` keeps `rounding` `pub(crate)`, so threshold can't call it. Exporting the scalar `decompose` and making `decompose_coefficient` a thin wrapper removes the copy. The exhaustive `[0, Q)` equivalence test does mitigate divergence risk, but duplicated fixed-point crypto constants across crates is the worst place in the workspace to carry a copy. (Related pre-existing instance: `make_hint_single` at `primitives.rs:539` independently reimplements MakeHint.)

While there: the `decompose_coefficient` doc (`primitives.rs:208-216`) still says it returns "(a₀ + q, a₁)" / "unsigned a₀ + q", but the code returns a₀ **mod q** (`+q` only when negative). That is precisely the convention change this PR is careful to document everywhere else — the doc on the function itself shouldn't misstate it.

## Verified clean — security / constant-time / memory

The things the review specifically went hunting for and did **not** find:

- **Constant time.** Every variant-dependent branch introduced by the genericization is on a *const generic*, folded at monomorphization — never on data. The γ₂=(Q−1)/32 arms of `decompose`/`make_hint`/`use_hint` are byte-identical to master; the new (Q−1)/88 arm is the reference-C branchless form. `rej_eta` η=4 (`4 − nibble`, accept `< 9`) reuses the same branchless masked-store/clamp machinery; `z_pack`/`z_unpack` 2^17, `eta_pack`/`eta_unpack` η=4, and the 6-bit `w1_pack` all match the reference bit layouts with no data-dependent branches. `check_norm`/`is_norm_within_bound` keep the full-sweep, OR-accumulated form. The `signature_var` rejection loop is structurally identical to master: all four bound checks always evaluated via non-short-circuiting `&`, the only leak channel remains the attempt count, and the nonce guard is now correctly derived from the variant's L.
- **Memory/zeroization.** `Polyvec<N>` keeps the `Zeroize, ZeroizeOnDrop` derives that `Polyveck`/`Polyvecl` had, so every drop-wipe in the sign loop is preserved; all explicit wipes in `keypair_var`/`public_key_from_secret_var`/`SigningContext` survived 1:1. The `define_ml_dsa!` frontend is semantically identical to the audited `ml_dsa_87.rs`: `Zeroizing` staging in `from_bytes`, `to_bytes → Zeroizing<[u8; N]>`, `sk.zeroize()` in `generate`, no `Clone` on key types, `from_parts` re-derivation. No new sk/seed copies. hdwallet hygiene is unchanged and the 87 path is an exact factoring of master's (the unchanged 5.4k-line golden-vector suite passes against it).
- **η=4 canonicality (the one check that couldn't transfer mechanically from 87).** `eta_unpack::<ETA>`'s range sweep correctly rejects non-canonical 4-bit slots 9–15 for ML-DSA-65 (and 3-bit slots 5–7 for η=2), accepts slot 8 (−4), keeps the non-short-circuiting fold, and the `#[must_use]` flag is still enforced at the single `public_key_from_secret_var` chokepoint reached by every import path in all three frontends.
- **Parameters.** All three variants match FIPS 204 Tables 1/2 exactly, including derived packed sizes (1312/2560/2420, 1952/4032/3309, 2592/4896/4627) and challenge entropy. Top-level re-exports stay ML-DSA-87; no 87 constant leaks into shared code paths. Packing buffer sizes are pinned by `const` assertions, so a mis-sized frontend cannot compile.
- **Threshold invariants.** Feature priority (87 > 65 > 44) is encoded identically at every cfg site with `compile_error!` when none is enabled — no cross-variant constant mixing is possible. Suite IDs are distinct. The κ invariant holds arithmetically (spot-checked radii × κ for every κ>1 entry, 44 and 87). 44's (4,6) reshare genuinely fails closed (guard `Err` → abort, pinned by a correctly-cfg'd test), and missing table entries fail closed everywhere (`resharing_kappa` defaults to 1.0, the strictest value). Per-config frame budgets, round-2 exact-length checks, and η-bounded deserialization all read active-variant constants (including η=4 on 65). The 87 tables and wire format are byte-identical to master.
- **Bot findings triaged.** The two CodeQL "hard-coded nonce" alerts (`poly.rs:1957/2033`) are fixed seeds inside `#[cfg(test)]` deterministic unit tests — false positives. Cursor's no_std-CI gap is already fixed in the current revision (explicit `--features ml-dsa-44,ml-dsa-65,ml-dsa-87` on the wasm jobs).

## Minor

1. **Zeroization regression tests only bind the 87 monomorphization.** `import_stack_zeroization.rs` / `keygen_zeroization.rs` assert *codegen* behavior (move elision, no seed-bearing frees), which is per-monomorphization — 44/65's different frame sizes can spill differently. Cheap to stamp per variant.
2. **Adversarial import tests live only in `ml_dsa_87.rs`** (mismatched pk, corrupted tr/t0, out-of-range coefficients, zero t1). In particular there's no end-to-end forged-blob test at η=4, the one canonicality case new in this PR (mitigated by the `eta4_unpack_rejects_noncanonical_slots` unit test + the shared chokepoint).
3. **hdwallet test targets break without `std`**: `tests_variants.rs:19/28/176` use `to_string()`/`String::new()` without alloc imports; `cargo check -p qp-rusty-crystals-hdwallet --tests --no-default-features --features ml-dsa-44,ml-dsa-65` fails (E0433/E0599). Latent since CI always enables `std` — gate the test modules on `std` or import from `alloc`.
4. **44's bound comment is inverted**: `MAX_COMMITMENT_DATA_SIZE` on 44 is 4196·2944 + 600 000 = 12 953 024, which *exceeds* `MAX_SIGNING_MESSAGE_SIZE` (12 582 912) — the docs claim headroom the other way. Legit frames still fit (worst case 12 353 065), but the effective cap silently becomes the global one with ~1.8% margin.
5. **Script/table sync hazard**: `compute_hyperball_params.py` keeps `(4,6): 1.25` in `RESHARING_DATA[44]["kappa"]` even though the shipped table intentionally omits it (fail-closed) — regenerating from the script can reintroduce the enlargement the crate declares infeasible. `SHIPPED_TABLES` also fully duplicates the Rust tables behind only a "keep in sync" comment.
6. **`SUBSHARE_COEFF_BOUND = 500` justification is η=2-specific**; it still holds at η=4 with ~5× margin, but the derivation text should cover 65.
7. Worth keeping visible for auditors: 65's aggregated split noise is ~3.3× below its η=4 keygen variance, so the "reshared shares look freshly dealt" hiding argument is quantitatively weaker there. It's documented in-code and in SECURITY_PROOF.md — good — flagging it here so it doesn't get lost.

## DRY inventory

Net-net this PR is a large *de*duplication — the eight `k_*`/`l_*` polyvec function pairs collapsed into single generics, the k_iterations table that existed twice on master now lives once in `params.rs`, and magic sizes (4627, 2592, 8·736…) are replaced by params re-exports throughout. Beyond required item 2, the residual duplication, roughly by value of fixing:

- `dilithium/src/frontend.rs`: `sign`/`verify` duplicate the full 14-const turbofish in their `Some(ctx)`/`None` arms (4 × ~16 lines) — hoisting the prefix construction leaves one call each. `Keypair::from_bytes` also re-implements `from_parts`' derive-and-compare instead of delegating.
- The 87>65>44 cfg-priority chain is hand-written at 6 sites in threshold (`params.rs` ×2, `mldsa.rs` ×2, bench, integration test). The two out-of-crate copies exist only because the active frontend module isn't publicly re-exported; a `pub use` plus one dispatch macro pins the priority in one place.
- `dilithium/src/params.rs`: the 8-line derived-constant block (`C_DASH_BYTES` … `SIGNBYTES`) is textually identical in all three variant modules — the base constants are data (fine); the derivation formulas are logic and could be stamped once.
- `ml_dsa_44.rs` vs `ml_dsa_65.rs` test modules are byte-identical except three size constants (~55 lines × 2); hdwallet's `variant_tests!` macro is the in-repo pattern to copy. Their `rejects_zero_t1_public_key` is also a third variation of 87's.
- `dilithium/benches/dilithium_benchmarks.rs`: three per-variant blocks identical except module path/label (~20 lines × 3) — a small macro collapses each to one line.
- `dilithium/examples/ct_bench.rs`: the `fill_polyvecl`/`fill_polyveck` and `copy_polyvecl`/`copy_polyveck` pairs now have literally identical bodies over `Polyvec<L>`/`Polyvec<K>` — this PR's own const-generic change makes each pair one function.
- Smaller: `sign.rs` recomputes `beta = TAU * ETA` in three functions though each params module defines `BETA`; 87's threshold tables are inline in `params.rs` while 44/65 are `include!`-files (asymmetric for regeneration); hdwallet's wormhole path re-inlines the `derive_entropy_from_seed` sequence the PR just factored out; the CI 44/65 test steps are copy-pasted modulo the digit.
- Stale 87-era docs on now-generic code: `dilithium/src/params.rs:8-10` ("current single-variant frontend is ML-DSA-87"), `threshold/src/protocol/primitives.rs:1` ("K = 8, L = 7"), `signing.rs:418` (ν values), `signer.rs:412`, and the 87-centric κ narrative above the per-variant paragraph in `resharing/protocol.rs`.

None of the DRY items are blocking on their own; the required list is just #1 and #2 above.

dilithium/src/frontend.rs
```diff
@@ -0,0 +1,790 @@
...
+				let public =
+					PublicKey::from_bytes(public_bytes).map_err(|_| KeyParsingError::BadKeypair)?;
+
+				// The SecretKey is constructed directly (not via
+				// `SecretKey::from_bytes`) because `from_parts` performs the
+				// same validation — both run the packed-invariant checks
+				// inside `public_key_from_secret_var` — and going through the
+				// import path first would redo that keygen-scale derivation.
+				// On failure the secret is wiped by its own drop.
+				Self::from_parts(SecretKey { bytes: *secret_bytes }, public)
```
n13: **[P1] Preserve stack zeroization across this delegation**

Dereferencing `secret_bytes` produces a plain secret-key array, and passing the resulting `SecretKey` by value through `from_parts` leaves a dead caller-frame copy that `SecretKey`'s drop cannot wipe. The new optimized probe finds this residue for all three variants on both CI platforms and locally, after the returned keypair itself has been zeroized. The pre-follow-up implementation, which validates by borrowing `secret_bytes` and then constructs the result directly, passes the same probe with the same toolchain. Please avoid this by-value hop—for example, factor the shared validation into a borrow-only helper and construct the `Keypair` directly—and keep all three release probes green.

n13 (changes requested):
## Verdict

**Request changes.**

The two blockers from my prior review—DKG/contribution domain separation and the cross-crate `decompose` duplication—are fixed cleanly, and the nonblocking follow-ups are mostly solid.

One new security regression blocks merge: the recent `Keypair::from_bytes` → `from_parts` dedup leaves a plaintext copy of the imported secret key in dead stack memory in every parameter-set monomorphization. Both required workspace jobs fail on macOS and Linux, and I reproduced the same 0/3 failure locally with:

`cargo test --release --locked -p qp-rusty-crystals-dilithium --all-features --test import_stack_zeroization`

Rebuilding revision `e882994` with the same compiler makes the original ML-DSA-87 probe pass (1/1), isolating the regression to the follow-up refactor rather than the probe or runner environment. This needs to be fixed and the three optimized probes kept green before merge.

Nonblocking documentation note: `threshold/src/protocol/primitives.rs:27-34` still says the module uses hard-coded ML-DSA-87 parameters (`K=8`, `L=7`, `ETA=2`), although it now selects the active variant.

n13 (changes requested):
## Review

Genericizing an audited ML-DSA-87 implementation over all three FIPS 204 parameter sets without disturbing the 87 path is a hard thing to do well, and the cryptographic core here is done well. The const-generic design (parameters as const generics + `const { assert!(...) }` blocks that reject inconsistent `(K, L, ETA, GAMMA1, GAMMA2, OMEGA, size)` tuples at monomorphization) means a mis-instantiated variant fails to compile rather than producing subtly wrong bytes. The validation story — exhaustive oracle tests, vendored ACVP, vendored NIST KATs, per-variant threshold calibration with the invariant that κ > 1 implies enlarged radii — is stronger than what the code had before.

One blocking item, one question I'd like answered, and a note for whoever drives the next release.

### What I verified locally (branch `ml-dsa-multi-variant`, `4938e2f`)

| Check | Result |
|---|---|
| `cargo check --workspace --all-targets` | clean |
| `cargo test --release --workspace` (default / 87) | **685 passed, 0 failed** |
| dilithium `--all-features`: ACVP keyGen/sigGen/sigVer × 44/65/87 | pass |
| `rounding::decompose` + `use_hint` exhaustive over `[0, Q)`, both γ₂ | pass |
| NIST KATs, 100 vectors each, Dilithium2/3/5 → ML-DSA-44/65/87 | pass (incl. exact signature reproduction under the AES-CTR DRBG) |
| threshold, `--no-default-features --features std,ml-dsa-44` | 260 lib + 49 integration + 65 resharing + 48 other, **all pass** |
| threshold, `--no-default-features --features std,ml-dsa-65` | 259 lib + 49 integration + 67 resharing + 48 other, **all pass** |
| release stack-zeroization probes, all three frontends | pass |
| clippy `-D warnings`, threshold + hdwallet at 44 and 65 | clean |
| clippy `-D warnings`, **dilithium/hdwallet with no `ml-dsa-*` feature** | ❌ **25 errors** (see below) |

The `test_resharing_into_4_of_6_fails_closed` guard on ML-DSA-44 does fire as documented.

---

## Blocking

### `--no-default-features` dilithium no longer passes clippy — and it's a regression

```
$ cargo clippy -p qp-rusty-crystals-dilithium --no-default-features -- -D warnings
error: unused macro definition: `define_ml_dsa`     --> dilithium/src/frontend.rs:35:14
error: function `ctx_prefix` is never used          --> dilithium/src/frontend.rs:16:15
error: function `keypair_var` is never used         --> dilithium/src/sign.rs:62:15
... 25 errors total
```

On `master` this command is clean. With no `ml-dsa-*` feature the entire signing core becomes dead code.

This directly contradicts a claim in the PR description — *"`std` only (no ML-DSA): wormhole/mnemonic surface builds and passes"* with *"clippy `-D warnings` clean"*. It does not; `cargo clippy -p qp-rusty-crystals-hdwallet --no-default-features --features std -- -D warnings` fails with the same 25 errors from the path dependency. That combination matters because the workspace pins dilithium with `default-features = false`, so it's a configuration real consumers land in.

CI misses it because the clippy job runs `--all-features`. Suggested fix: a crate-level `#![cfg_attr(not(any(feature = "ml-dsa-44", feature = "ml-dsa-65", feature = "ml-dsa-87")), allow(dead_code))]`, and add the featureless build to the clippy matrix so it stays fixed.

---

## Note for the next release (not this PR)

Next releases should be 4.0.0 due to many breaking changes


---

## Security-relevant — I'd like an answer on the first one, but neither blocks

### ML-DSA-65's resharing hiding argument is knowingly weaker, and nothing in the API says so

`SPLIT_NOISE_NUM_X256 = 125` is tuned so aggregated split noise σ ≈ σ_keygen for **η = 2**. ML-DSA-65 uses η = 4 (keygen variance 20/3), so on 65 the split noise sits ~3.3× *below* keygen variance. The PR is admirably honest about this — it's documented at the constant, in `SECURITY_PROOF.md` §Parameter-Set Scope, and in `resharing_norm_enlargement` — and correctly notes it is *norm*-safe (overshoots 0.62–0.92, so κ = 1 everywhere).

But "norm-safe" and "private" are different claims. The a-posteriori *"reshared shares are distributed like a fresh keygen sharing"* argument is what §Privacy leans on, and on 65 it is quantitatively weaker by an amount the PR doesn't bound. The asymmetry is uncomfortable: a caller who picks 65 over 87 to save bandwidth gets a resharing privacy argument that's weaker than the one they'd get at either η = 2 set.

I'd like one of:
- a concrete distinguishing-advantage bound for 65 resharing, or
- 65 resharing behind an explicit opt-in feature / `#[doc(cfg)]`-level "experimental" marker, so nobody reaches it by accident.

Signing on 65 is unaffected — this is resharing-only.

### No constant-time coverage for the new η = 4 sampler path

`ct_bench` and `polyvec::ct_internals` are still hard-pinned to `params::ETA` (= 2, ML-DSA-87). The η = 4 arm of `rej_eta` has a different acceptance probability (9/16 vs 15/16) and therefore a different rejection-count distribution, and `uniform_eta` runs 3 fixed rounds instead of 2 — a genuinely new secret-sampling path with **zero** dudect coverage. The reasoning in the `uniform_eta` doc comment for why 3 rounds suffices (~14σ) is sound and I don't dispute it; I'd just like the harness to be parameterized so it's measured rather than argued.

### `SUBSHARE_COEFF_BOUND = 500` is shared across variants

The doc honestly recomputes the honest-behavior estimate for η = 4 (~90 vs ~46) and notes the margin drops from ~10× to ~5×. Fine as shipped, but it's now the one resharing bound that *isn't* derived from `params` — consider moving it into the per-variant tables alongside κ so it can't drift.

### Deserialization bounds grew on the "cheaper" parameter sets

`MAX_COMMITMENT_DATA_SIZE` 10.5 MB → **~12.95 MB** on 44 (`MAX_K_ITERATIONS` = 4196 at (5,6) × 2944 + 600 kB slack); `MAX_RESPONSE_SIZE` 8 MB → ~10.47 MB. So the *lowest* security level admits the *largest* untrusted frame. The PR documents the cost, and the const-assert in `signing_protocol.rs` correctly pins the worst-case frame under `MAX_SIGNING_MESSAGE_SIZE`.

Two follow-ups worth considering:

- These are **global worst-case** bounds. A deployment fixed at (2,3) on ML-DSA-44 (k = 10, ~29 kB of real commitment data) still lets an attacker push a ~13 MB frame through the bounded deserializer before the per-config `max_frame_size` check runs. Where the session config is known at parse time, tightening to `config.k_iterations() × SINGLE_COMMITMENT_SIZE` would be cheap hardening.
- The const-assert margin on 44 is **~1.9%**: the ceiling is `MAX_K_ITERATIONS ≤ 4274` and the table ships 4196. Fail-loud (compile error) is the right behavior, but a comment stating that ceiling would save the next person a confusing build break after a recalibration.

### 44/65 tables are calibrated but not audited

Called out clearly in the PR. Worth carrying into the crate docs too, so a downstream reading `README.md` sees that 87's tables carry an audit and 44/65's carry a 2,000-sample grid search.

---

## Correctness review — no defects found

I checked the parts most likely to hide a variant-specific bug:

- **Parameters.** FIPS 204 Table 1 values are correct for all three sets (τ 39/49/60, λ 128/192/256, γ₁ 2¹⁷/2¹⁹/2¹⁹, γ₂ (q−1)/88 ÷ /32 ÷ /32, (k,ℓ) (4,4)/(6,5)/(8,7), η 2/4/2, ω 80/55/75), and the derived sizes are pinned against Table 2.
- **`rounding::decompose`, γ₂ = (Q−1)/88 arm.** The `(11275, >> 24)` reciprocal and the `a1 ^= ((43 - a1) >> 31) & a1` branchless wrap match the PQCrystals reference exactly. Both arms are checked exhaustively over all 8.4 M inputs against an oracle written straight from Algorithm 36, plus reconstruction and `use_hint` against Algorithm 40.
- **Codecs.** The new `eta_{pack,unpack}` 4-bit arm, `z_{pack,unpack}` 18-bit arm, and `w1_pack` 6-bit arm all match the reference bit layouts. The canonicality sweep in `eta_unpack` stays correct for η = 4 (slots 9..15 decode to −5..−11, still `< -ETA`).
- **`uniform_eta` round count.** Raising `fixed_rounds` 2 → 3 for η = 4 is necessary — at 9/16 acceptance, two blocks would fall short of 256 at ~1e-5, an observable timing variation. Correctly identified and correctly justified.
- **The `decompose_coefficient` convention change** (always `a₀ + q` → `a₀ mod q`) is safe. `combine_signature` is the only consumer that reads `w0`, and it reduces `w0 + difference` through `mod_q` before use; `generate_round3_response` computes `w0` and never reads it. The exhaustive test pins mod-q equivalence against the audited 87 original kept verbatim as an oracle — exactly the right way to land this kind of change.
- **Domain separation** is thorough and consistent: signing SSID v2 → v3 absorbing version + suite, DKG SSID V1 → V2, DKG contribution v2 → v3, resharing suite ID tracking the active set. Each has a negative test that a substituted suite ID changes the digest.
- **Constant-time properties survive genericization.** Every variant branch (`GAMMA2`, `ETA`, `GAMMA1` arms) is a constant condition folded at monomorphization, not a data-dependent branch. `use_hint`'s real branches run only on public verification data, as documented.

---

## Nits

- `config.rs::unsupported_config_reason()` `match`es on the `VARIANT_NAME` *string* to pick between three near-identical `&'static str`s. It works around `InvalidParameters { reason: &'static str }`, but a `pub const UNSUPPORTED_CONFIG_REASON` next to `VARIANT_NAME` in each `active` module would be direct and wouldn't restate the variant name in a third place.
- `sign::verify_var` takes `ETA`, `PK`, and `SK` const params it never uses, purely to feed `assert_sign_params`. Splitting that assert into sign-side and verify-side halves would drop three parameters from every verify call site (which are already 14 lines of turbofish).
- `frontend::ctx_prefix` never writes `buf[0]` — it relies on the caller passing a zeroed buffer. Both call sites do, but that's an unstated precondition on a function that builds a domain separator; either write `buf[0] = 0` or state it in the doc.
- `generate_round3_response` allocates and fills a `w0` that is never read.
- `primitives::POLY_Q_SIZE` is now just an alias for `params::POLY_Q_PACKEDBYTES`; one of the two should go.
- `run_all_tests.sh` covers the dilithium feature matrix but not the 44/65 threshold/hdwallet builds that CI now runs — easy to drift.
- Repo weight: +4.4 MB of vendored ACVP JSON, +1.8 MB of `.rsp` KATs, +2.2 k lines of vendored hdwallet vectors. Worth confirming that's an accepted trade for offline reproducibility.

---

## Verdict

**Request changes** — on one mechanical item: fix the `--no-default-features` dead-code failure and add that configuration to the clippy matrix, so the build config the workspace itself uses stays green and the PR's testing claim holds.

The ML-DSA-65 resharing privacy question I'd like an answer on — either a bound or an experimental gate — but I don't think it needs to block the merge, given 65 resharing is opt-in by feature and the caveat is documented in three places. The major-version note is for the release, not for this PR.

The cryptographic work is solid and well-validated, and I'd be glad to approve once the clippy regression is fixed.

n13 (approved):
## Verdict

**Approve.**

The one blocker from my last review is fixed exactly as suggested and verified locally, and the follow-ups were addressed beyond what I asked — most notably, ML-DSA-65's resharing hiding gap was closed by recalibrating the split noise to full keygen parity rather than just documenting the asymmetry.

### What I verified locally (branch `ml-dsa-multi-variant`, `c65d432`)

| Check | Result |
|---|---|
| clippy `-D warnings`, dilithium `--no-default-features` (the blocker) | **clean** (was 25 errors) |
| clippy `-D warnings`, hdwallet `--no-default-features --features std` | clean |
| clippy `-D warnings`, `--workspace --all-targets --all-features` | clean |
| `cargo test --release --workspace` (default / 87) | all green |
| threshold `std,ml-dsa-65` exclusive (release) — 260 lib + 67 resharing on the **recalibrated κ tables** + all integration suites | all green |
| threshold `std,ml-dsa-44` exclusive (release) — 261 lib + 65 resharing incl. the (4,6) fail-closed pin | all green |
| hdwallet `std,ml-dsa-44` / `std,ml-dsa-65` exclusive | all green |

### Disposition of the prior round's items

1. **Featureless clippy regression (blocking)** — fixed with the crate-level `cfg_attr` allow plus two new CI clippy steps for exactly the configurations that were missed (`analysis` job). The CI comment explaining *why* `--all-features` can't catch this is a nice touch.
2. **ML-DSA-65 resharing hiding** — resolved the strong way. The split-noise numerator is now per-committee (`split_noise_num_x256(S_old)`) and calibrated so the fixed-point conditional variance is ≥ 1.02× the η = 4 keygen variance (measured 1.028–1.051×), restoring the a-posteriori "fresh keygen sharing" argument to the same strength as the η = 2 sets. I reviewed the mechanism closely:
   - **The η = 2 path is bit-for-bit unchanged**: `desired ≤ 125 < 127` forces `num_draws = 1` with the same rounding as the old `split_noise_threshold`, so the audited 87 (and 44) resharing PRF streams are untouched. Confirmed by inspection and by the green 87/44 suites.
   - **Stream alignment holds**: the multi-draw sampler consumes exactly `num_draws` PRF bytes per coefficient, and `num_draws` is deterministic in `(ETA, S_old)`, so all members of an old subset still derive identical sub-shares.
   - **The κ invariant holds**: (2,4)/(3,5)/(4,6) radii are scaled by exactly 1.10/1.10/1.18 (I checked the arithmetic against the base radii), `K` re-derived at the enlarged radii (14/208/7560, the (4,6) entry at 400k MC samples), the variance regressions assert overshoot ≤ κ from the same table the guard reads, and the new `#[ignore]`d multi-seed test pins the thin (4,6) margin (1.131 worst-of-6 vs κ = 1.18) across independent keygen seeds.
   - The cost is honest and documented: 65's (4,6) K = 7560 means ~33.4 MB Round-2 frames, and the const assert pins the fit under the raised cap.
3. **η = 4 dudect coverage** — the harness is now stamped per η via macro (`keygen_s1s2_sampling_eta{2,4}`, `rej_eta_one_block_eta{2,4}`), with `ct_internals` genericized as passthroughs. The example stays gated on `ct-internals` and the `params::ml_dsa_*` const modules are ungated, so there's no feature-combination compile hazard. (Verified it builds under `--all-targets --all-features`; I did not run a full dudect measurement session.)
4. **`SUBSHARE_COEFF_BOUND`** — now const-derived from the active η with a unit test pinning both arms (450 at η = 2, 890 at η = 4). Note this *tightens* the 87/44 receive-side bound from 500 to 450 — fine (honest max ≈ 46, so the 10× margin is intact, and nothing is deployed), but it does touch a validation constant on the audited path; flagging so it's a conscious change.
5. **Nits** — `UNSUPPORTED_CONFIG_REASON` via `concat!` ✓, `ctx_prefix` writes its own zeros ✓, `POLY_Q_SIZE` alias removed ✓, `run_all_tests.sh` mirrors the CI 44/65 matrix ✓, the `w0` slot in `generate_round3_response` is now documented as a required output of the shared helper ✓.

### Non-blocking follow-ups

- **hdwallet README version drift**: the version revert to 3.0.1 updated line 22 to `"3.0"` but lines 19 and 45 still say `"3.1"`. One-line fix whenever convenient.
- **`MAX_SIGNING_MESSAGE_SIZE` is a flat 36 MiB on every variant**: it's sized by 65's (4,6) worst case, so 87/44 builds now admit ~3× larger pre-parse frames than their own worst case (~9.4/12.4 MB). Since `MAX_COMMITMENT_DATA_SIZE` is already const-derived per variant, this cap could be too (e.g. `FRAME_HEADER_LEN + MAX_COMMITMENT_DATA_SIZE` rounded up) — same family as the per-config tightening note from last round.
- **The η = 4 noise table has no drift pin**: `protocol.rs::split_noise_num_x256` is duplicated in `keyhiding_conditional.py`, and the keygen-parity evidence rests on the script reproducing the Rust splitter exactly. The crate already pins script↔table agreement for the hyperball tables via `include_str!`; the same trick would cover the NUM table.
- Cosmetic: stray blank lines after the (4,6) rows in `params_tables_65.rs`.

### Bot findings on this PR

The three CodeQL "hard-coded cryptographic value / nonce" alerts are in `#[cfg(test)]` code and the dudect bench harness — fixed seeds/nonces are deterministic by design there; safe to dismiss. Cursor's no_std finding is already fixed on this branch (the wasm job enables all three frontends explicitly).

The release-versioning note from last round stands: the next dilithium release should be a major (4.0.0) given the public-API changes; leaving manifests for CI-driven bumping is fine as long as the automation majors it.

Solid work across all the rounds — the calibration discipline (measure, pin, fail closed) held up under every change. Ship it.

PR #107, opened 2026-07-30
V12 all variants

illuzen:
## Summary

Security-review follow-ups for threshold resharing / parameter validation, plus related secret-hygiene and public rounding API hardening.

### Resharing liveness & config fail-closed

- **Reject unsupported new committees at construction** (`784b23b`): ML-DSA-44 declares reshares *into* a 4-of-6 committee infeasible (no κ enlargement; Round-5 recovered-partial guard rejects honest reshares at κ = 1). `resharing_kappa()` still returned `1.0` for missing table rows, and `ResharingConfig::new()` only checked signing support via `K_ITERATIONS`, so sessions could burn Rounds 1–5 before a guaranteed late failure. Each parameter table now ships `RESHARING_UNSUPPORTED`; `params::resharing_supported()` and `ResharingConfig` reject those new committees up front (`UnsupportedNewCommittee`). Freshly-dealt 4-of-6 signing remains valid.

- **Expected active set is an inclusion policy** (`6bccbf9`): `set_expected_active_set()` only accelerated the Act proposal; the leader still built Act from *all* Round-1 commitments. An old member left out of the expected set could send one Ready packet, enter Act, then go silent — a one-packet liveness veto. When an expected set is configured, Act is now capped to expected ∩ committed members.

- **Drop malformed frames from non-required senders** (`222d01b`): `message()` returned hard `MalformedMessage` for garbage from any old∪new participant before active-set filtering. An excluded old member could abort every handoff under runners that treat `message()` errors as fatal. Escalation is gated on `session_depends_on()` (new committee always; old members only once in Act); everyone else is logged and dropped.

### Dilithium / threshold API contracts

- **Canonicalize signed representatives in public rounding** (`7d0a69c`): `Poly::from_coeffs` admits `(-Q, Q)`, but `rounding::{power2round,decompose}` assumed `[0, Q)`, so valid signed inputs mis-rounded (e.g. `power2round(-1) ≠ power2round(Q-1)`). Both helpers now `caddq` first; docs state the domain. `make_hint` stays branchless (docs-only precondition).

- **Fail-closed participant mask conversions** (`e27e1a4`): `mask_from_ids` / `ids_from_mask` silently dropped unknown IDs and out-of-range bits, shrinking a requested quorum with no signal. Both return `Option` and reject unrepresentable inputs (aligned with `subset()`). The resharing recovered-partial error path maps an impossible out-of-range generated mask to `InternalError` (no `.expect`).

### Related hygiene (same branch)

- Keep secret-dependent norm diagnostics out of Round-5 wire messages (`d473839`).
- Clarify working-copy vs durable old-share ownership in resharing docs (`fb80e38`).
- Wipe secret-bearing stack temporaries outside zeroizing owners (`2c9fbf6`).

## Test plan

- [ ] `cargo test --locked -p qp-rusty-crystals-threshold --no-default-features --features std,ml-dsa-44 --lib` (fail-closed 4-of-6 reshare config + tables)
- [ ] `cargo test --locked -p qp-rusty-crystals-threshold --no-default-features --features std,ml-dsa-65 --lib` / default (87) for `resharing_supported` / 4-of-6 still accepted where κ ships
- [ ] `cargo test --release --locked -p qp-rusty-crystals-threshold --test resharing_tests` (expected-Act inclusion, malformed-frame drop, existing handoff suite)
- [ ] `cargo test --release --locked -p qp-rusty-crystals-dilithium --all-features` (rounding canonicalization + ACVP / zeroization)
- [ ] `cargo test --locked -p qp-rusty-crystals-threshold --lib participants::` (mask fail-closed + doctests)
- [ ] Confirm CI exclusive-feature jobs for `ml-dsa-44` / `ml-dsa-65` still green

---

> [!NOTE]
> **High Risk**
> Changes touch threshold protocol liveness, quorum selection, and cryptographic rounding invariants where mistakes could enable DoS, undersized quorums, or incorrect decompositions; mitigated by extensive regression tests and mostly fail-closed behavior.
> 
> **Overview**
> Security-review follow-ups across **threshold resharing**, **dilithium rounding**, and **secret hygiene**.
> 
> **Resharing** now rejects impossible new committees at `ResharingConfig::new` via per-variant `RESHARING_UNSUPPORTED` and `params::resharing_supported()` — e.g. ML-DSA-44 reshares *into* 4-of-6 would always fail Round 5, so coordinators no longer burn Rounds 1–5. `set_expected_active_set` is an **inclusion cap**: `Act` is `expected ∩ committed`, blocking a one-packet liveness veto from an old member left out on purpose. Malformed wire bytes only become hard `MalformedMessage` when `session_depends_on` the sender; excluded old members’ garbage is dropped. Round 5 public failure reasons stay fixed (no secret-dependent norm diagnostics on the wire). Docs clarify **working copy vs durable** old share.
> 
> **Dilithium** `power2round` / `decompose` call branchless `caddq` first so signed `(-Q, Q)` inputs match their standard representatives; `ParticipantList` mask conversions return `Option` and fail closed on unknown IDs or out-of-range mask bits.
> 
> **Hygiene**: zeroize signing/HD-wallet/DKG/resharing stack copies and SHAKE squeeze buffers; document why `keccakf1600` round locals are not individually wiped.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e27e1a46fe7ea9002dfdf40a64d8403095f3f2fa. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


n13 (approved):
## Verdict

**Approve on substance — but CI is red on `cargo fmt`, and that failure skipped every test job, so nothing in the test plan actually ran on CI.** Fix the formatting and let CI go green before merge. I ran the test plan locally at e27e1a4 and everything passes.

## Overview

Security-review follow-ups in four areas:

1. **Fail-closed reshare targets** — per-variant `RESHARING_UNSUPPORTED` tables + `params::resharing_supported()`; `ResharingConfig::new` rejects committees the Round-5 guard is guaranteed to reject (ML-DSA-44 4-of-6) with `UnsupportedNewCommittee` instead of burning Rounds 1–5.
2. **Expected active set as inclusion cap** — `Act` is now `expected ∩ committed`, closing the one-packet liveness veto from an old member deliberately left out of the expected set.
3. **Malformed-frame trust boundary** — garbage bytes only escalate to a hard `MalformedMessage` when `session_depends_on(sender)` (new committee always; old members once in Act); everyone else is logged and dropped, removing a one-packet abort surface under runners that treat `message()` errors as fatal.
4. **Rounding canonicalization + hygiene** — `power2round`/`decompose` `caddq` first so the full `Poly::from_coeffs` domain `(-Q, Q)` rounds correctly; mask conversions return `Option` and fail closed; Round-5 wire errors carry a fixed, value-free reason; assorted stack-copy zeroization in sign/hdwallet/DKG.

## Verified locally (CI never got to the test jobs)

- `cargo check --workspace --all-features` — clean.
- dilithium lib tests, all features, release: **153 passed** (incl. the new canonicalization tests and the exhaustive `decompose`/`use_hint` sweeps).
- threshold lib tests — default/ML-DSA-87: **263 passed**; ML-DSA-44: **265 passed** (fail-closed 4-of-6 config + `RESHARING_UNSUPPORTED` consistency tests).
- `resharing_tests` targeted: `active_set*` and `malformed*` all pass.
- hdwallet tests: all pass.

Spot-checks on the risky bits:

- `caddq` is branchless and maps `(-Q, 0)` → `(0, Q)` correctly. Both internal callers of the rounding helpers (`power2round` in keygen, `decompose` in signing) already run `polyvec::caddq` beforehand, so this changes **no produced keys or signatures** — it only fixes the public contract for `Poly::from_coeffs`-range inputs. The docs/`debug_assert` split for out-of-contract inputs is sensible.
- `expected_active_set` is sorted + deduped at set time, so the filtered `Act` stays sorted and `is_active`'s `binary_search` stays valid. `|Act| ≥ t_old` is enforced on every proposal trigger path (fast path, expected-complete, window-close).
- Validation order in `ResharingConfig::new` is right: `ThresholdConfig::new` (K_ITERATIONS membership) runs first, so `UnsupportedNewCommittee` only fires for signing-valid-but-unresharable committees.
- `session_depends_on` sits after the self/non-participant filters; new-committee and Act members still get attributable hard errors (accountability preserved, covered by the new test).
- The only internal `ids_from_mask` caller (recovered-partial error path) was updated; `mask_from_ids` has no internal callers. The `InternalError` mapping instead of `.expect` matches the crate's fail-early style.
- Zeroization changes: `poly.rs` buffers are plain arrays with `Zeroize` impls, the hdwallet HMAC output wipe happens after the halves are consumed, and the `keccakf1600` non-wiping rationale is a reasonable, honestly documented boundary.

## Required before merge

- **`cargo fmt`** — the Format job fails on 4 diffs: `dilithium/src/poly.rs:1294`, `threshold/src/params.rs:182` (the tab-indented `#[cfg(test)]`), `threshold/src/resharing/protocol.rs:2242`, `threshold/src/resharing/types.rs:1903`. Everything downstream (tests, clippy, no_std, stack budget, audit) was skipped because of it.

## Non-blocking notes

- `ids_from_mask`/`mask_from_ids` returning `Option` is a **breaking change** to the threshold crate's public API (plus the new `ResharingConfigError` variant), and `threshold/Cargo.toml` is still `2.0.0` on this branch. Assuming the release automation handles the major bump as with the recent dilithium/hdwallet bumps — just flagging so it isn't published as a patch.
- Pre-Act, a garbled frame from a member of the *expected* set is now silently dropped (`session_depends_on` is false until Act is agreed), so a leader waiting on that member stalls to the ready-window timeout instead of failing fast with blame. That's the right trade-off — the alternative reopens the abort surface — but be aware the failure diagnosis moved from a hard error to a `log::warn` + timeout.
- `make_hint` keeping a docs-only precondition (no runtime check) is correct given its constant-time purpose; the new doc making the FFI-caller obligation explicit is the best available option.

PR #108, opened 2026-07-31
Zeroization and further hardening

illuzen:
# Security review fixes: zeroization, validation, and protocol-contract enforcement

This PR addresses 13 security review findings across `hdwallet`, `dilithium`, and `threshold`. Every fix follows the same discipline: reproduce the finding with a failing (red) test where testable, fix it, and keep the test as a regression pin. All items were verified as real before fixing; one reviewer recommendation was deliberately amended based on empirical measurement (see the last item).

## Secret zeroization and memory hygiene (`hdwallet`, `dilithium`)

- **HMAC-SHA512 state leaked HD-derivation secrets** (`ecc9fc2`). The `hmac`/`sha2` crates don't wipe their internal buffers on drop, leaving the BIP-39 seed, parent secret keys, and chain codes in freed memory. Replaced with an in-crate HMAC-SHA512 that zeroizes all intermediate state, verified against reference test vectors and painted-stack probes.
- **Dropped `Mnemonic`s retained the phrase** (`4f9f113`). Enabled the `bip39` crate's `zeroize` feature so mnemonics are wiped on drop.
- **BIP-39 seeds returned by value** (`6c4d5cc`). `mnemonic_to_seed` returned a raw `[u8; 64]`, so every caller had to remember to wipe it — and Rust's move semantics left unwiped stack copies even when they did. The seed is now written into a caller-owned `SensitiveBytes64` out-parameter, filled in place.
- **`ExtendedPrivKey` accessors and derivation materialized unguarded key copies** (`35ebccd`). `secret()` now returns `&SensitiveBytes32` instead of a raw copy; `derive()` fills a caller-owned out-parameter and child steps happen in place, eliminating the intermediate by-value returns that left key material on the stack.
- **Keypair/wormhole generation entropy crossed function boundaries by value** (`aa6c6c3`). `Keypair::generate` and `WormholePair::generate_new` now take `&mut SensitiveBytes32` and wipe it after use; `keypair_var` borrows instead of consuming.
- **`t0_pack`/`eta_pack` staging buffers held secret-derived values** (`66e3666`). The packing helpers now zeroize their local staging arrays; release-mode painted-stack tests pin the behavior against future codegen changes.

All the stack-residue fixes are verified by painted-stack probe tests asserting zero secret residue after the operation.

## Deserialization and construction validation (`threshold`)

- **`PublicKey::new` skipped validation** (`9a6b1b0`). The internal constructor accepted bytes that `from_bytes`/Borsh would reject (e.g. the degenerate all-zero t1 encoding). It is now fallible and validates on every construction path; DKG, resharing, and the trusted dealer propagate the error, and `pack_combined_pk` rejects degenerate combined keys.
- **`PrivateKeyShare` import accepted structurally invalid share maps** (`c327747`). Deserialization now verifies the share map keys exactly equal the holder's canonical subset-mask set, rejecting missing, extraneous, and non-holder masks at the import boundary instead of failing rounds later at signing recovery.

## Protocol lifecycle and contracts (`threshold`)

- **Fatal DKG errors left the session in a retryable-looking round** (`eb7a4fa`). Every fatal error from an active phase now routes through a common abort path that zeroizes all secret state (round state, seed, pending and buffered private messages) and lands in the terminal `Failed` phase. `is_failed()`/`is_complete()` are now public.
- **Signing constructor accepted any leader** (`cdd2413`). `DilithiumSignProtocol::new` now enforces the documented leader-election rule (lowest ID in the signing set), turning silent leader disagreement — which would surface as a Round 4 stall — into an immediate configuration error.
- **`PrivateKeyShare::Clone` had no documented duplication policy** (`2ee58d7`). The type-level docs now spell out that clones are independent plaintext copies with their own zeroization lifecycle and full custody obligations; `Clone` is implemented manually so the policy and the secret-field list are explicit, and a test pins the independent-lifecycle semantics.

## Documentation contract (`hdwallet`)

- **README contradicted the hardened-only parser** (`6be3852`). The README claimed `change`/`address_index` may be unhardened while the parser deliberately rejects any unhardened segment. Hardened-only is correct for lattice keys (no public-key homomorphism, mirroring SLIP-10's ed25519 rule); the README now documents that contract and the parser test pins both sides of it.

## Resharing norm guard (`threshold`) — one deliberate deviation from the review

- **Stored-share norm guard was ~7.75× looser than intended** (`cfa77cf`). The Round 5 stored-share guard compares a *raw* (unchallenged) share norm, but its bound reused the recovered-partial helper, which bakes in the `√τ` challenge-amplification factor — accepting individual stored shares far outside the single-share envelope while public-key and recovered-partial checks still passed. The bound is now the raw envelope `1.3·√(dim·var_η)·κ`, with `√τ` applied explicitly at the recovered-partial call site so the convention difference is visible.

  The review recommended dropping κ as well; measurement says otherwise. Over 20 chained reshares, honest per-share inflation reaches ~1.24× on (4,6) — nearly the full 1.3 base headroom — so without κ honest reshares would sit one bad draw from rejection. With κ the worst honest norm/bound ratio is 0.83, while the envelope is still `√τ ≈ 7.75×` tighter than before.

## Breaking API changes

Callers of the following need updating:

| API | Before | After |
|---|---|---|
| `hdwallet::mnemonic_to_seed` | returns `[u8; 64]` | fills `&mut SensitiveBytes64` |
| `ExtendedPrivKey::secret()` | returns `[u8; 32]` | returns `&SensitiveBytes32` |
| `ExtendedPrivKey::derive` | returns by value | fills `&mut ExtendedPrivKey` |
| `Keypair::generate` / `WormholePair::generate_new` | consume entropy by value | take `&mut SensitiveBytes32`, wipe after use |
| `threshold::PublicKey::new` | infallible | fallible (validates bytes) |
| `DilithiumSignProtocol::new` | any member as `leader_id` | `leader_id` must be the lowest ID in `participants` |

## Testing

- Red-then-green regression test for every code fix; painted-stack probes (release mode) for the zeroization items.
- Full workspace test suites green; clippy clean.
- The tightened resharing norm bound was validated empirically against 2,040 honest stored shares from steady-state chained reshares across all supported committee shapes.

---

> [!NOTE]
> **High Risk**
> Changes secret-handling and threshold protocol boundaries (DKG abort, leader rule, share-map and public-key validation) with several breaking public APIs; regressions are broad but mistakes could affect key derivation or distributed ceremonies.
> 
> **Overview**
> **Security-review hardening** across `hdwallet`, `dilithium`, and `threshold`, with red/green and painted-stack (release) regression tests.
> 
> **Secret lifecycle (`hdwallet` / `dilithium`)** replaces `hmac`-crate derivation with in-crate **HMAC-SHA512 / PBKDF2** that wipes hash state; enables **`bip39` `zeroize`**; writes BIP-39 seeds and HD keys via **caller-owned `SensitiveBytes64` / `ExtendedPrivKey` out-parameters** instead of by-value returns; changes **`Keypair::generate`**, **`keypair_var`**, and **`WormholePair::generate_new`** to take **`&mut SensitiveBytes32`** and zeroize in place; adds **`SensitiveBytes32/64::zeroed()` / `as_mut_bytes()`**; zeroizes **`t0_pack` / `eta_pack`** staging buffers.
> 
> **Threshold** makes **`PublicKey::new` / Borsh import** validate like `from_bytes` and rejects **degenerate combined keys** in DKG/dealer/resharing; **`PrivateKeyShare` Borsh** must match the holder’s **canonical subset-mask set**; fatal **DKG `poke` errors** call **`abort()`** (wipe secrets, **`Failed` phase**, **`is_failed` / `is_complete`**); **`DilithiumSignProtocol::new`** requires **`leader_id`** to be the **lowest** participant; documents and tests **`PrivateKeyShare::Clone`** as independent plaintext copies; tightens **resharing stored-share norm bound** (drops erroneous **√τ** on raw norms, keeps **κ**).
> 
> **Docs / contract:** `hdwallet` README now states **hardened-only** paths (aligned with the parser).
> 
> **Breaking callers:** `mnemonic_to_seed`, `ExtendedPrivKey::derive` / `secret()`, entropy-taking keygen/wormhole APIs, fallible `PublicKey::new`, and signing leader validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cfa77cfc8669c05a964a8eb77bb332820d06c631. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,6 +88,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "90dbd31c98227229239363921e60fcf5e558e43ec69094d46fc4996f08d1d5bc"
 dependencies = [
  "bitcoin_hashes",
+ "zeroize",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -838,6 +838,7 @@ dependencies = [
  "hex",
  "hex-literal",
  "hmac",
+ "psm",
  "qp-poseidon-core",
  "qp-rusty-crystals-dilithium",
  "rand",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,9 @@ resolver = "2"
 
 [workspace.dependencies]
 bincode = { version = "=1.3.3", default-features = false }
-bip39 = { version = "=2.2.2", default-features = false }
+# `zeroize` is required: without it, dropped `Mnemonic` values leave their
+# word-index buffer (from which the phrase is reconstructible) in dead memory.
+bip39 = { version = "=2.2.2", default-features = false, features = ["zeroize"] }
 borsh = { version = "=1.6.1", default-features = false, features = ["derive"] }
 criterion = { version = "=0.5.1", features = ["html_reports"] }
 getrandom = { version = "=0.2.17", features = ["js"] }
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This implementation provides enterprise-grade memory security:
 // Secure by design - entropy is zeroized after conversion
 let mut entropy = [0u8; 32];
 getrandom::getrandom(&mut entropy).unwrap();
-let keypair = ml_dsa_87::Keypair::generate((&mut entropy).into());
+let keypair = ml_dsa_87::Keypair::generate(&mut (&mut entropy).into());
 // entropy is now [0,0,0,...] - no sensitive data left in memory
 
 // If you genuinely need a second copy of a keypair, you must opt in explicitly:
@@ -56,7 +56,7 @@ let mut entropy = [0u8; 32];
 getrandom::getrandom(&mut entropy).expect("Failed to generate entropy");
 
 // Generate keypair
-let keypair = ml_dsa_87::Keypair::generate((&mut entropy).into());
+let keypair = ml_dsa_87::Keypair::generate(&mut (&mut entropy).into());
 
 // Sign message
 let message = b"Hello, post-quantum world!";
--- a/dilithium/README.md
+++ b/dilithium/README.md
@@ -40,7 +40,7 @@ use qp_rusty_crystals_dilithium::ml_dsa_87;
 // Generate a keypair with secure random entropy
 let mut entropy = [0u8; 32];
 getrandom::getrandom(&mut entropy).expect("Failed to generate entropy");
-let keypair = ml_dsa_87::Keypair::generate((&mut entropy).into());
+let keypair = ml_dsa_87::Keypair::generate(&mut (&mut entropy).into());
 
 // Sign a message
 let message = b"Hello, post-quantum world!";
@@ -58,7 +58,7 @@ use qp_rusty_crystals_dilithium::ml_dsa_87;
 
 let mut entropy = [0u8; 32];
 getrandom::getrandom(&mut entropy).expect("Failed to generate entropy");
-let keypair = ml_dsa_87::Keypair::generate((&mut entropy).into());
+let keypair = ml_dsa_87::Keypair::generate(&mut (&mut entropy).into());
 
 let message = b"Important message";
 let context = b"email-signature-v1"; // Domain separation
@@ -75,7 +75,7 @@ use qp_rusty_crystals_dilithium::ml_dsa_87;
 
 let mut entropy = [0u8; 32];
 getrandom::getrandom(&mut entropy).expect("Failed to generate entropy");
-let keypair = ml_dsa_87::Keypair::generate((&mut entropy).into());
+let keypair = ml_dsa_87::Keypair::generate(&mut (&mut entropy).into());
 
 let message = b"Message to sign";
 
--- a/dilithium/benches/dilithium_benchmarks.rs
+++ b/dilithium/benches/dilithium_benchmarks.rs
@@ -25,14 +25,14 @@ macro_rules! bench_ml_dsa {
 		#[cfg(feature = $feature)]
 		{
 			use qp_rusty_crystals_dilithium::$mod_name::Keypair;
-			let keypair = Keypair::generate((&mut [2u8; 32]).into());
+			let keypair = Keypair::generate(&mut (&mut [2u8; 32]).into());
 			let msg = b"";
 			let sig = keypair.sign(msg, None, None).expect("Signing should succeed");
 			bench_variant(
 				$c,
 				$label,
 				|| {
-					let _ = Keypair::generate((&mut [1u8; 32]).into());
+					let _ = Keypair::generate(&mut (&mut [1u8; 32]).into());
 				},
 				|| {
 					let _ = keypair.sign(msg, None, None);
--- a/dilithium/examples/ct_bench.rs
+++ b/dilithium/examples/ct_bench.rs
@@ -210,7 +210,7 @@ fn sign_sk_expansion(runner: &mut CtRunner, rng: &mut BenchRng) {
 	let gen_sk = |rng: &mut BenchRng| -> [u8; params::SECRETKEYBYTES] {
 		let mut entropy = [0u8; 32];
 		rng.fill_bytes(&mut entropy);
-		let keypair = Keypair::generate((&mut entropy).into());
+		let keypair = Keypair::generate(&mut (&mut entropy).into());
 		// Bench-only throwaway keys: copy out of the self-wiping buffer.
 		*keypair.secret().to_bytes()
 	};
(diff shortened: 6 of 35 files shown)


n13 (approved):
# Review

**Verdict: Approve** — the substance is sound and independently verified; two mechanical CI items need fixing before merge.

## What I verified (not just re-read)

- **Ran the full release test suites locally** (`hdwallet`, `dilithium`, `threshold`, integration tests): all green, including every new painted-stack probe. This matters because CI's test job never ran on this PR (skipped due to the format failure below).
- **Leader-election enforcement** (`signing_protocol.rs`): `ParticipantList::new` sorts its input, so `participant_list.get(0)` really is the lowest ID. The check also correctly subsumes the old membership check.
- **DKG abort path** (`keygen/dkg/protocol.rs`): every "not ready" case in the round handlers returns `Ok(DkgAction::Wait)`, so routing all `Err`s through `abort()` is safe — no transient error gets promoted to a terminal failure. `abort()` wipes the same field set as the existing `Drop` impl, and the wipe-then-set ordering correctly handles `DkgState::zeroize` resetting `phase`/`error_message`.
- **Share-map completeness check** (`keys.rs` Borsh import): matches what producers actually emit — the dealer inserts exactly the masks containing the holder's bit (`dealer.rs:328`), so honest blobs from all existing producers still import. `MAX_PARTIES = 6` keeps the `1u16 << my_dkg_index` shift and the subset enumeration trivially safe.
- **Norm-bound refactor** (`resharing/protocol.rs`): algebraically checked — the recovered-partial bound is unchanged (`√τ` moved out of the shared envelope and reapplied at the call site), and the stored-share bound is tightened by exactly `√τ`. Keeping κ is well argued by the honest-inflation measurement.
- **In-crate HMAC-SHA512/PBKDF2** (`hderive.rs`): pinned by RFC 4231 vectors, cross-checked against the `hmac` crate across key lengths and message-part splits, and the BIP-39 path is cross-checked against `bip39::to_seed_normalized`. NFKD passphrase normalization is retained. The `String::with_capacity(mnemonic.len())` reasoning for the canonical phrase holds (every word is a substring of the parsed input and separators are ≥1 byte, so no realloc can leave an unwiped heap copy).

## Required before merge

1. **CI is red on formatting**: nightly-rustfmt drift in `threshold/src/resharing/protocol.rs` and `threshold/src/signing_protocol.rs` (comment rewrapping + one `expect_err` reflow — all from nightly-only options like `wrap_comments`, which is why stable `cargo fmt --check` passes locally). Run `cargo +nightly fmt`. This failure also skipped the test/clippy/audit/no_std/stack-budget jobs, so none of them have run on this PR.
2. **CodeQL "high" alert is a false positive** — `hdwallet/src/tests.rs:767`, "cleartext logging of sensitive information" on `assert!(..., "got {err:?}")` for `PassphraseTooLong`. The variant carries only a `usize` byte length, and it's test code. Dismiss it in the code-scanning UI (it will keep blocking the check otherwise).

## Minor (non-blocking)

- `hmac_sha512`'s one-block key limit is only a `debug_assert!` — in release, an oversized key silently produces a non-RFC-2104 MAC instead of failing. All current callers are safe (`pbkdf2_hmac_sha512` pre-hashes long passwords), but a hard `assert!` would fail early if a future caller misses that contract.
- `probe_stack_for` now exists in five near-identical copies (`dilithium/src/poly.rs`, `dilithium/tests/import_stack_zeroization.rs`, three new `hdwallet/tests/*` files). The three hdwallet integration tests could share one helper via `tests/common/mod.rs`.
- `Dkg::abort` duplicates the `Drop` impl's wipe list. A shared `fn wipe_secrets(&mut self)` called from both would prevent a future secret field from being added to one and missed in the other.
- The new packing probes cover `t0_pack` and `eta_pack::<2>` but not the η=4 branch (ML-DSA-65), whose staging temporaries are scalar locals — almost certainly register-allocated, but that's the same class of codegen assumption these probes exist to pin. A matching probe would be cheap.
- The breaking-changes table omits `ExtendedPrivKey::child()` → `child_in_place()` (and `child_in_place` is now infallible — a nice simplification).
- Observation only: `pbkdf2_hmac_sha512` recomputes the HMAC key schedule every iteration (~2× the SHA-512 compressions of a precomputed-state PBKDF2). At 2048 rounds this is ~1 ms — simplicity is the right trade here.

## Overall

This is a high-quality hardening pass. The red-test-first discipline shows: every behavioral claim in the description maps to a pinned regression test, the painted-stack probes are correctly gated to release builds with self-checks that prove the detection technique works, and the one place the PR deviates from the reviewer's recommendation (keeping κ) is justified with actual measurement rather than argument. The out-parameter API shape is unidiomatic Rust, but the doc comments explain *why* move semantics force it, which is exactly what a future maintainer needs to not "clean it up" back into a leak.

n13 (approved):
Lol Fable called Copilot's false positive

GTG

PR #112, opened 2026-08-03
More zeroization

illuzen:
# Security review round: timing, zeroization, protocol liveness, and API hardening

Works through a batch of security-review findings across `dilithium`,
`hdwallet`, and `threshold`. Every accepted finding got a regression test
first (red), then a fix (green); findings we assessed as invalid or
already-mitigated are documented where the reviewer would look next.

## Depends on

**`qp-poseidon` PR "Wipe the sponge state after hashing"** (branch
`illuzen/zeroization`). This branch temporarily points `qp-poseidon-core` at
that local branch via `[patch.crates-io]` — the wormhole stack probe needs
the hasher to stop leaving digest copies in dead stack frames. Before merge,
replace the patch with the released `=3.0.3` pin.

## Constant-time comparisons

- `WormholePair::eq` compared secret bytes with short-circuiting `==`, a
  timing oracle against the secret. All three fields now compare through a
  branchless `ct_eq_32` primitive (exposed as `SensitiveBytes32::ct_eq`),
  combined with non-short-circuiting `&`. Covered by a `dudect` bench and a
  unit test pinning that every field participates in equality.

## Zeroization: painted-stack probes and the leaks they caught

New release-mode regression tests paint a stack buffer (`psm`), run an
operation on it, and scan for secret residue afterwards. Rust moves are
copies that leave the moved-from slot dead but never dropped, so
`ZeroizeOnDrop` alone cannot reach them — these probes pin the codegen.

Probes added: trusted dealer, DKG (subset shared secret), threshold signing
Round 1 (mask randomness), dilithium keygen + signing, `SensitiveBytes`
constructors, wormhole pair generation, and the derived-key contribution
digest. A heap-scanner scenario for the dealer was added to the existing
allocator-hook test.

Leaks found and fixed:

- **Dealer**: per-party seeds sat in a plain `Vec` and share coefficients in
  unwiped locals; the dead non-NTT aggregate `s1_total` was removed outright.
- **Dilithium signing/keygen**: `unpack_secret_key_for_signing` returned the
  unpacked key by value (dead copy of `K`); `Keypair::generate` left two dead
  copies of the packed secret key. Both now use out-parameter /
  tail-construction shapes.
- **`SensitiveBytes{32,64}` constructors**: returning the named local left
  the secret in the constructor's frame; they now `mem::replace` out of a
  zeroed slot.
- **Wormhole**: the secret crossed four by-value boundaries. `into_bytes` is
  removed from `SensitiveBytes` (extracting the inner `Copy` array recreates
  every hazard the wrapper prevents), the unused `WormholePair::verify` is
  removed, and derivation now borrows the secret and swaps it into the pair.
- **Derived-key contribution**: the 64-byte secret-share digest (which
  reproduces a party's DKG contribution for any tweak) was returned by value
  and held in a plain local; now squeezed into a caller-provided `Zeroizing`
  buffer.

## Threshold protocol hardening

- **Round 3 intake gate**: wrong-length Round 3 responses used to occupy the
  sender's first-wins slot and only surface later as an unattributable
  `combine()` failure. They are now rejected at receive time.
- **DKG Round 4 liveness**: a malicious party could abort the session after
  all four rounds by appending one extra valid-but-non-led subset mask to its
  broadcast. Verification now skips subsets the sender does not lead, exactly
  as the aggregation step would.
- **`DkgConfig::new`** now rejects a local signer whose public key does not
  match the key registered for `my_party_id` — previously this misconfig
  surfaced as peers rejecting the Round 4 transcript signature.
- **Key-import DoS**: `from_bytes` ran keygen-scale public-key re-derivation
  before the cheap coefficient range check; the order is now reversed so
  garbage inputs are rejected early.
- **`poly_pack_w`** OR-ed bits into its buffer, silently relying on callers
  to pre-zero it; it now clears the buffer itself.

## Documentation corrections

- **Transport contract**: `DilithiumSignProtocol::message` now documents that
  `from` must be the transport-authenticated identity, and why unauthenticated
  transports degrade to DoS only.
- **DKG integrity model**: partial-key η-binding requires an honest subset
  member; documented as a secure-with-abort known limitation (module docs and
  keygen README) rather than papered over.
- **`derivation.rs`** claimed the legacy `key` field is derived from public
  inputs and "provides no secrecy" — inverted since the v2 derivations bind
  it to the secret shares. Docs now mark it secret (never log or expose) and
  keep the correct rationale for not binding contributions to it: pre-v2
  shares carry a public-derived `key`, so the polynomials are the only
  binding sound for every share vintage.

## API hardening

- `WormholePair` fields are now private with read-only accessors
  (`address()`, `first_hash()`, `secret()`), making the
  "address/first_hash are derived from secret" invariant structural. No
  external consumer touched the fields directly.

## Testing

- Full workspace suites pass in both debug and release (probes are
  release-only by design).
- KATs and golden vectors unchanged throughout — no derivation, hash, or
  wire-format behavior changed except the documented intake rejections.

---

> [!NOTE]
> **High Risk**
> Touches secret handling, signing/keygen, DKG, and threshold wire behavior across the crypto stack; wrong zeroization or intake logic could leak keys or change protocol completion, though KATs/golden vectors are reported unchanged aside from documented rejections.
> 
> **Overview**
> This PR implements a security-review pass across **dilithium**, **hdwallet**, and **threshold**, with regressions added before each fix. It temporarily **`[patch.crates-io]`** `qp-poseidon-core` to a local branch that wipes sponge state after hashing (needed for wormhole stack probes); that patch should be dropped once **=3.0.3** is released.
> 
> **Constant-time and sensitive APIs:** Adds **`ct_eq_32`** and **`SensitiveBytes32::ct_eq`**; **`WormholePair::PartialEq`** now folds all fields with non-short-circuiting `&`. **`SensitiveBytes`** constructors use in-place fill + **`mem::replace`**; **`into_bytes`** is removed. **`WormholePair`** fields are private with accessors; **`verify`** is removed in favor of re-derivation.
> 
> **Zeroization:** Signing/keygen and wormhole paths avoid by-value moves of secrets (out-parameters, tail construction, borrow-and-swap). Dealer holds party seeds in **`Zeroizing<Vec<_>>`** and wipes share copy locals; DKG contribution digest is written into a caller **`Zeroizing`** buffer. Release-only **painted-stack** tests (`psm`) cover dealer, DKG, signing Round 1, dilithium keygen/signing, sensitive-byte constructors, and wormhole generation.
> 
> **Threshold protocol:** Rejects malformed **Round 3** responses at intake; **DKG Round 4** ignores partial PKs for subsets the sender does not lead; **`DkgConfig::new`** fails if the local signer key mismatches the registered map entry; secret-key import runs a cheap coefficient range check before keygen-scale re-derivation; **`poly_pack_w`** zeroes its output buffer.
> 
> **Docs:** Transport contract for **`DilithiumSignProtocol::message`** (`from` must be authenticated); DKG partial-key η-binding documented as secure-with-abort; derivation docs corrected for the **`key`** field and share vintage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e41dd9a56abff4e1fa8a15d5a3b7f98dff8557f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -814,8 +814,6 @@ dependencies = [
 [[package]]
 name = "qp-poseidon-core"
 version = "3.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "55e32db7b3a5d70086f82f198ecfc021f17cb7ec70a416512dba3488f4a116c1"
 
 [[package]]
 name = "qp-rusty-crystals-dilithium"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -862,6 +862,7 @@ dependencies = [
  "hex-literal",
  "libm",
  "log",
+ "psm",
  "qp-rusty-crystals-dilithium",
  "rand",
  "zeroize",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -823,6 +823,7 @@ dependencies = [
  "dudect-bencher",
  "hex",
  "psm",
+ "qp-rusty-crystals-test-utils",
  "rand",
  "serde_json",
  "zeroize",
@@ -840,6 +841,7 @@ dependencies = [
  "psm",
  "qp-poseidon-core",
  "qp-rusty-crystals-dilithium",
+ "qp-rusty-crystals-test-utils",
  "rand",
  "rand_core",
  "serde",
@@ -848,6 +850,13 @@ dependencies = [
  "thiserror",
  "unicode-normalization",
  "zeroize",
+]
+
+[[package]]
+name = "qp-rusty-crystals-test-utils"
+version = "0.1.0"
+dependencies = [
+ "psm",
 ]
 
 [[package]]
@@ -862,6 +871,7 @@ dependencies = [
  "log",
  "psm",
  "qp-rusty-crystals-dilithium",
+ "qp-rusty-crystals-test-utils",
  "rand",
  "zeroize",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -813,7 +813,9 @@ dependencies = [
 
 [[package]]
 name = "qp-poseidon-core"
-version = "3.0.2"
+version = "3.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5872607e25ea4ee5fb37e64bf1462168e1a36a4e719cdc8a105533c708253918"
 
 [[package]]
 name = "qp-rusty-crystals-dilithium"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["dilithium", "hdwallet", "tests", "test-utils", "threshold"]
+members = ["dilithium", "hdwallet", "test-utils", "tests", "threshold"]
 resolver = "2"
 
 [workspace.dependencies]
@@ -17,8 +17,8 @@ log = { version = "=0.4.29", default-features = false }
 qp-poseidon-core = { version = "=3.0.2", default-features = false }
 qp-rusty-crystals-dilithium = { path = "./dilithium", version = "4.0.0", default-features = false }
 qp-rusty-crystals-hdwallet = { path = "./hdwallet", version = "4.0.0", default-features = false }
-qp-rusty-crystals-threshold = { path = "./threshold", version = "=3.0.0", default-features = false }
 qp-rusty-crystals-test-utils = { path = "./test-utils" }
+qp-rusty-crystals-threshold = { path = "./threshold", version = "=3.0.0", default-features = false }
 rand = { version = "=0.10.1", default-features = false, features = ["std_rng", "thread_rng"] }
 rand_core = { version = "=0.10.0", default-features = false }
 rusty-crystals-integration-tests = { path = "./tests" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["dilithium", "hdwallet", "tests", "threshold"]
+members = ["dilithium", "hdwallet", "tests", "test-utils", "threshold"]
 resolver = "2"
 
 [workspace.dependencies]
@@ -18,6 +18,7 @@ qp-poseidon-core = { version = "=3.0.2", default-features = false }
 qp-rusty-crystals-dilithium = { path = "./dilithium", version = "4.0.0", default-features = false }
 qp-rusty-crystals-hdwallet = { path = "./hdwallet", version = "4.0.0", default-features = false }
 qp-rusty-crystals-threshold = { path = "./threshold", version = "=3.0.0", default-features = false }
+qp-rusty-crystals-test-utils = { path = "./test-utils" }
 rand = { version = "=0.10.1", default-features = false, features = ["std_rng", "thread_rng"] }
 rand_core = { version = "=0.10.0", default-features = false }
 rusty-crystals-integration-tests = { path = "./tests" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ hex = { version = "=0.4.3", default-features = false, features = ["alloc"] }
 hex-literal = { version = "=0.4.1", default-features = false }
 libm = { version = "=0.2.11" }
 log = { version = "=0.4.29", default-features = false }
-qp-poseidon-core = { version = "=3.0.2", default-features = false }
+qp-poseidon-core = { version = "=3.1.0", default-features = false }
 qp-rusty-crystals-dilithium = { path = "./dilithium", version = "4.0.0", default-features = false }
 qp-rusty-crystals-hdwallet = { path = "./hdwallet", version = "4.0.0", default-features = false }
 qp-rusty-crystals-test-utils = { path = "./test-utils" }
@@ -27,11 +27,3 @@ serde_json = { version = "=1.0.150", default-features = false, features = ["allo
 thiserror = { version = "=2.0.18", default-features = false }
 unicode-normalization = { version = "=0.1.25", default-features = false }
 zeroize = { version = "=1.8.2", default-features = false, features = ["derive"] }
-
-# TEMPORARY (security review): test against the local qp-poseidon branch that
-# wipes the sponge state after hashing (`illuzen/zeroization`), so hash inputs
-# and outputs — wormhole secrets are both — stop lingering in dead stack
-# frames. Remove once that branch is released as =3.0.3 and the pin above is
-# bumped.
-[patch.crates-io]
-qp-poseidon-core = { path = "../qp-poseidon" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,3 +26,11 @@ serde_json = { version = "=1.0.150", default-features = false, features = ["allo
 thiserror = { version = "=2.0.18", default-features = false }
 unicode-normalization = { version = "=0.1.25", default-features = false }
 zeroize = { version = "=1.8.2", default-features = false, features = ["derive"] }
+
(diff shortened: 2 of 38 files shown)


n13 (changes requested):
## Review

### Overview

A security-review batch across `dilithium`, `hdwallet`, and `threshold`: constant-time comparisons for wormhole secrets, painted-stack/heap regression probes plus the zeroization fixes they caught, threshold protocol hardening (Round 3 intake gate, DKG Round 4 liveness, config-time signer check, import DoS ordering, `poly_pack_w` self-zeroing), doc corrections, and `WormholePair` field privatization. Test-first discipline is visible throughout, and the PR description matches the diff accurately.

### Verified against the base branch

- **Round 3 intake gate** exactly mirrors what `unpack_responses` enforces (`k_iterations * L * POLYZ_PACKEDBYTES`, `threshold/src/protocol/signing.rs:694-704`), so no honest message can be dropped and every dropped message was structurally guaranteed to fail `combine()`. Both the direct and buffered paths are tested.
- **Round 4 non-led-subset skip** is consistent with aggregation: `collect_and_verify_all_partial_pks` only ever inserts PKs where `config.is_leader(subset)` holds, so ignoring non-led entries closes the abort vector without skipping any verification that mattered. The regression test additionally pins that the tampered view aggregates the same public key as honest parties.
- **`!s_in_range` early return** in `public_key_from_secret_var` wipes exactly the set the fall-through path wipes (`key`, `s1`, `s2`, `t0`; `t0_derived` does not exist yet at that point), and the timing rationale is sound — the range of an imported blob is known to whoever supplied it.
- **`DkgConfig` signer/registered-key check** compiles against the existing `PublicKey: Clone + PartialEq` bound (`types.rs:77`); no bound changes needed.
- **Dealer cleanup**: the removed `s1_total` was genuinely dead (returned as `_s1_total`, never consumed). `party_keys` is pre-allocated with exact capacity, so the `Zeroizing<Vec<_>>` block is the only heap block the seeds ever land in — no realloc can strand a stale buffer.
- KATs and golden vectors are untouched; observable wire behavior changes only in the documented intake rejections.

The test methodology deserves a call-out: probes self-check that the technique detects a deliberately planted leak, patterns are recomputed independently through public APIs so the harness cannot plant a match, and reference values are captured outside the probed region. That is the right way to build this class of test.

### Blocking

1. **`[patch.crates-io]` path dependency** (`qp-poseidon-core = { path = "../qp-poseidon" }`) and the matching `Cargo.lock` entry with no source/checksum. As written, the build requires a sibling checkout of an unmerged `qp-poseidon` branch — CI's workspace build cannot pass, and neither can a fresh clone. Already flagged in the PR description as temporary; per the repo's release discipline this should merge only after `qp-poseidon-core` `=3.0.3` is live on crates.io and the pin is bumped, with the patch section and lockfile change dropped.
2. **CI is red on formatting**: `cargo fmt` diffs in `threshold/src/keygen/dkg/protocol.rs` (~line 2009) and `threshold/src/keygen/dkg/types.rs` (~line 1115). Because Fast Checks failed, every downstream job (workspace tests, clippy, no_std, stack budget, audit) was skipped — the full matrix has not actually run on this head commit.

### Non-blocking suggestions

- **Test-helper duplication**: `probe_stack_for` now exists in six near-identical copies (2× dilithium tests, 1× hdwallet, 2× threshold tests + 1 inline in `derivation.rs`), and `dealer_party0_seed_pattern` / `hyperball_stream_pattern` are each duplicated between `heap_zeroization.rs` and their new stack-probe siblings. At minimum the three threshold copies could share a `tests/common/mod.rs`; cross-crate sharing (a dev-only util crate) is worth considering if more probes are coming.
- **Breaking public API changes** — `SensitiveBytes{32,64}::into_bytes` removed, `WormholePair` fields privatized, `WormholePair::verify` removed — are semver-major for `dilithium` and `hdwallet` (and transitively `threshold`). Fine given the release automation, but worth stating explicitly in the release notes.
- **`first_hash()` returns `[u8; 32]` by value** while its own doc says it is "a reveal value, not public data". Mild tension with the copy-hygiene this PR enforces everywhere else; returning `&[u8; 32]` (as `secret()` does) would make disclosure explicit at call sites. Same applies to `address()`, though there it's harmless.
- **Round 3 rejection log** attributes the drop to `r3.party_id`, which is unauthenticated payload data; the PR's own transport-contract docs establish `from` as the identity to blame. Logging `from` (or both) would make the attribution trustworthy.
- **`ct_eq_32`** is hand-rolled rather than using `subtle`; given the dudect bench, the `black_box` barrier, and the workspace's pinned-dependency posture, that's a reasonable trade, just noting the conventional alternative.
- The painted-stack probes pin codegen shapes (`mem::replace`, tail construction, out-parameters) that the optimizer is not obligated to preserve, so expect occasional churn on rustc upgrades. The release-only `cfg` and the precedent already in the repo make that an accepted cost — a probe regressing on a toolchain bump is a signal to re-inspect, not necessarily a source bug.

### Verdict

**Request changes — on the two mechanical blockers only** (the fmt failure, and replacing the temporary `qp-poseidon` path patch with the released `=3.0.3` pin so CI can actually run the workspace suite). The substantive content — the fixes, the threat modeling in the docs, and the regression-test technique — is excellent, and I'd approve as-is once those land and CI is green.

n13 (approved):
## Re-review

All blockers and all of the non-blocking suggestions from my previous review are addressed. Reviewed the delta (`8e41dd9a...85ae1209`) commit by commit:

### Blockers — resolved

1. **`[patch.crates-io]` removed** (85ae1209): the workspace now pins `qp-poseidon-core = "=3.1.0"` from crates.io, `Cargo.lock` carries a proper registry source + checksum, and the test doc referencing `>= 3.0.3` was updated to `>= 3.1.0`. Dependency-resolving jobs (clippy/doc, audit, no_std, stack budget) are green, confirming 3.1.0 resolves from the registry.
2. **Formatting** (e28b1955): Fast Checks (Format) passes; the two `cargo fmt` diffs in `threshold/src/keygen/dkg/{protocol,types}.rs` are fixed.

### Suggestions — all taken, and taken well

- **Probe dedup** (b77648ec, e27ce327): new `publish = false` workspace crate `qp-rusty-crystals-test-utils` with `probe_stack_for` / `probe_stack_for_named`; all eight integration-test copies collapsed onto it, and the hderive multi-pattern variant was unified into `probe_stack_for_named` rather than kept as a parallel shape. The threshold pattern helpers (`dealer_party0_seed_pattern`, `hyperball_stream_pattern`) moved to `tests/common/stack_patterns.rs`, included via `#[path]` so the resharing/coverage suites don't see unused items. I checked each refactored test against its previous version: every scenario, self-check assertion, and pattern derivation is preserved verbatim — this is a pure mechanical consolidation, no probe was weakened.
- **Borrowed accessors** (179b2b2b): `address()` / `first_hash()` now return `&[u8; 32]`, making any copy an explicit `*` at the call site, consistent with `secret()`.
- **Log attribution** (179b2b2b): both the Round 3 gate *and* the pre-existing Round 2 rejection log now attribute to the transport-authenticated `from`, with the claimed `party_id` reported separately — better than what I asked for.

### One tiny nit (non-blocking, fine to leave)

The comment justifying the two remaining local `probe_stack_for` copies (`dilithium/src/poly.rs`, `threshold/src/derivation.rs`) says "unit tests cannot take a `dev-dependency`" — but dev-dependencies *are* available to unit tests (the `derivation.rs` probe itself calls `psm`, a dev-dependency, one line below that comment). The local copies are likely removable in a follow-up; not worth holding this PR for.

### Verdict

**Approve.** Note the two Test Workspace jobs (ubuntu/macos) were still running at review time — the release-mode probe suites are slow — so merge on green as usual; everything that has completed is passing.

PR #114, opened 2026-08-03
Security review round: secret custody on the heap, the stack, and the public API

illuzen:
# Security review round: secret custody on the heap, the stack, and the public API

Works through the latest batch of security-review findings. Every item was first
verified against the code; where a leak was reproducible a failing (red) test was
written before the fix, and every fix ships with a regression test that pins the
new behavior. Two findings were assessed as not exploitable under the documented
threat model and are addressed with documentation, logging, and pinning tests
instead of behavior changes.

### hdwallet: `generate_mnemonic` returned the recovery phrase as a plain `String`

The 24-word BIP-39 phrase is root secret material, but the API handed it to
callers with no erasure obligation, so it could persist in freed heap memory
after drop. It now returns `Zeroizing<String>`; the phrase is joined directly
into the single allocation the wrapper owns and wipes.

- **Red test:** a heap-scanning global allocator that greps freed blocks for the
  mnemonic pattern (`hdwallet/tests/mnemonic_heap_zeroization.rs`) — failed
  before, green after.
- **Breaking change:** callers now receive `Zeroizing<String>`; use `.as_str()`
  / deref. READMEs and tests updated.

### threshold: DKG `pending_privates` queue stranded K_S in freed `Vec` buffers

`Round1Private` is `ZeroizeOnDrop`, but `Vec` reallocation copies elements
byte-wise and frees the old buffer without wiping it, so queued subset secrets
K_S survived in dead heap blocks. Elements are now `Box`ed: growth moves only
pointers, and each secret has exactly one heap home that is zeroized in place on
drop.

- **Red test:** a 2-of-5 DKG under a freed-memory-scanning allocator
  (`threshold/tests/dkg_pending_privates_realloc_zeroization.rs`) — failed
  before, green after.

### threshold: DKG `round1_privates` buffer hardened against B-tree node splits

Same class of hazard as above: `BTreeMap` rebalancing copies entries byte-wise
between nodes. Values are now `Box`ed so node operations move pointers only. No
red test was possible — supported committee sizes keep the map below the node-
split threshold — so this one is defense-in-depth.

### dilithium: hedged signing passed the randomizer `rnd` as bare `Copy` bytes

The hedge crossed four by-value `Option<[u8; 32]>` parameter slots on the
signing chain with only the innermost copy zeroized. A recovered `rnd` plus a
compromised nonce seed `K` reconstructs `ρ' = H(K || rnd || μ)` and enables the
known-mask attack hedging exists to prevent. The hedge is now threaded as
`Option<&SensitiveBytes32>` end to end (matching `Keypair::generate`'s
convention): the caller's raw buffer is wiped when the wrapper is built, signing
only ever borrows the bytes, and deterministic mode absorbs a public zero block.

- **Test note:** the release-mode painted-stack probe gained a hedged scenario.
  It did *not* fail against the old code on the tested target (the optimizer
  happened to keep the copies out of dead stack memory), so this fix removes a
  codegen-dependent hazard and an API-custody gap rather than a demonstrated
  leak; the probe stays as a regression test. ACVP and NIST KAT vectors confirm
  signatures are byte-identical.
- **Breaking change:** `sign(msg, ctx, hedge)` now takes
  `Option<&SensitiveBytes32>`. README examples updated.

### threshold: resharing wire types exposed plaintext shares and the signer

`NewShareData.s1/.s2`, `ResharingRound4Message.contributions`, and
`ResharingSignerConfig.my_signer`/`.verifying_keys` were `pub`, letting
downstream code read, log, or retain plaintext sub-shares (and extract the
long-term transcript signer) outside their `ZeroizeOnDrop` lifecycle. All are
now `pub(crate)`, matching the `PrivateKeyShare` opacity convention;
`ResharingSignerConfig::new` (which all callers already used) is the only
construction path.

- **Pinned by:** `compile_fail` doctests on all three types (the red-test
  equivalent for API exposure — they compile, and thus fail, against the old
  code).
- The crate's own adversary-simulation tests keep forge/tamper access through
  `#[doc(hidden)]` helpers gated behind the non-default `internal-test-helpers`
  feature, enabled only for the crate's own test builds via a self-referential
  dev-dependency. Caveat: the types keep their public Borsh impls (wire
  format), so a determined consumer can still serialize to extract bytes — this
  closes the accidental path, not a hard boundary.

### dilithium: `polyvec::invntt_tomont` documented an unsafe precondition

The wrapper's docs promised inputs `< 2*Q` are fine, but the underlying
`poly::invntt_tomont` requires `|c| < Q`; coefficients in `[Q, 2Q)` can overflow
the i32 inverse butterflies in release builds and silently corrupt the
polynomial. In-crate callers already reduce below Q, so this was a contract fix:
docs corrected to the stricter bound, pinned by a `#[should_panic]` test on the
debug assertion.

### threshold: resharing first-message-wins buffering (Round 4 / accepts)

The review flagged that `handle_round4_message` and `handle_accept_message`
buffer the first message per sender without content verification, so a bad
first message occupies the slot. Under the crate's documented
sender-authenticated transport requirement, that scenario reduces to a sender
poisoning *its own* slot — no stronger than withholding (attributed abort and
restart) — while replace-on-duplicate would enable adaptive equivocation.
First-wins is therefore the anti-equivocation policy, kept and made explicit:
both handlers now document the argument, dropped duplicates are logged (warning
level when accept signatures conflict, which is observable equivocation), and a
regression test pins that duplicates never displace the first message.

## Testing

- Full workspace suite passes (unit, integration, doctests).
- Release-mode painted-stack probes pass for all three ML-DSA parameter sets.
- Heap-scanning allocator tests cover the mnemonic and DKG queue fixes.
- ACVP sigGen/keyGen/sigVer and NIST KAT vectors pass unchanged, confirming no
  cryptographic behavior drift.

---

> [!NOTE]
> **High Risk**
> Touches signing hedge custody, mnemonic return type, and threshold secret buffering with public API breaks; cryptographic behavior is pinned by ACVP/KAT and new zeroization tests.
> 
> **Overview**
> Security-review hardening across signing, HD wallet, and threshold protocols, with regression tests and **breaking API changes** where callers must adopt self-wiping wrappers.
> 
> **ML-DSA (dilithium):** Hedged signing now takes `Option<&SensitiveBytes32>` end-to-end instead of copying `rnd` through the call chain; README, ACVP, KAT, and integration tests updated. `polyvec::invntt_tomont` docs now require `|c| < Q` (not `< 2*Q`), with a debug `should_panic` test. Release stack probe adds a hedged-signing scenario.
> 
> **hdwallet:** `generate_mnemonic` returns `Zeroizing<String>` so the recovery phrase is wiped on drop; new heap-scanning test in `mnemonic_heap_zeroization.rs`.
> 
> **threshold:** DKG queues/buffers store `Round1Private` in `Box`es so `Vec`/`BTreeMap` growth does not free unwiped K_S; new 2-of-5 realloc heap test. Resharing makes sub-shares and signer config `pub(crate)` with `internal-test-helpers` for adversary tests; first-message-wins duplicate policy documented and pinned by unit test.
> 
> **Assessed (no behavior change):** Round 4 / accept buffering stays first-wins with logging on conflicting accepts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1713bcfcc3c635d0a16fd41b9ee4340a4e8a909a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,23 +86,49 @@ jobs:
             cargo test --locked -p qp-rusty-crystals-hdwallet --no-default-features --features "std,$variant"
             echo "::endgroup::"
           done
-      # The stack-zeroization guard is gated to optimized builds
+      # The stack-zeroization guards are gated to optimized builds
       # (#![cfg(not(debug_assertions))]): debug codegen materializes compiler
       # temporaries of the secret key that no source-level fix can wipe, so the
-      # debug `cargo test` above compiles this file to zero tests. The property
-      # it guards is entirely a function of codegen, making it the test most
-      # likely to silently regress on a toolchain bump — run it in release, and
-      # fail if zero tests ran so a gating/rename change can't turn the guard
-      # back into documentation.
+      # debug `cargo test` above compiles these files to zero tests. The
+      # property they guard is entirely a function of codegen, making them the
+      # tests most likely to silently regress on a toolchain bump — run every
+      # probe in release (security review: only the import probe used to run,
+      # leaving the sign/keygen/hedge and all threshold/hdwallet probes
+      # unexecuted by any job), and fail per probe if zero tests ran so a
+      # gating/rename change can't turn a guard back into documentation.
+      # Dilithium probes run --all-features, which stamps every ML-DSA
+      # parameter set into the binary; threshold/hdwallet probes run under the
+      # default (priority) set, matching what ships.
       - name: Optimized-build regression tests (stack zeroization)
         run: |
           set -o pipefail
-          cargo test --release --locked -p qp-rusty-crystals-dilithium --all-features \
-            --test import_stack_zeroization 2>&1 | tee stack-zeroization.log
-          grep -Eq "test result: ok\. [1-9][0-9]* passed" stack-zeroization.log || {
-            echo "::error::stack-zeroization tests were filtered out (0 tests ran)"
-            exit 1
+          run_probe() {
+            echo "::group::$*"
+            cargo test --release --locked "$@" 2>&1 | tee stack-zeroization.log
+            grep -Eq "test result: ok\. [1-9][0-9]* passed" stack-zeroization.log || {
+              echo "::error::stack-zeroization probe ($*) was filtered out (0 tests ran)"
+              exit 1
+            }
+            echo "::endgroup::"
           }
+          for probe in import_stack_zeroization sign_stack_zeroization sensitive_bytes_stack_zeroization; do
+            run_probe -p qp-rusty-crystals-dilithium --all-features --test "$probe"
+          done
+          for probe in dkg_stack_zeroization signing_stack_zeroization dealer_stack_zeroization; do
+            run_probe -p qp-rusty-crystals-threshold --test "$probe"
+          done
+          # Release-gated unit test inside the threshold lib (painted-stack
+          # probe over the derivation digest).
+          run_probe -p qp-rusty-crystals-threshold --lib shares_digest_never_survives_derivation
+          # The two mnemonic heap probes also run in the debug job, but the
+          # word-pointer scrub in generate_mnemonic is plain (safe) stores
+          # ahead of a free — exactly what dead-store elimination targets —
+          # so they must additionally pass under optimized codegen.
+          for probe in bip39_stack_zeroization entropy_stack_zeroization \
+              hderive_stack_zeroization wormhole_stack_zeroization \
+              mnemonic_heap_zeroization mnemonic_word_pointer_zeroization; do
+            run_probe -p qp-rusty-crystals-hdwallet --test "$probe"
+          done
 
   no-std-build:
     name: 📦 no_std Build (wasm32v1-none)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -874,6 +874,7 @@ dependencies = [
  "psm",
  "qp-rusty-crystals-dilithium",
  "qp-rusty-crystals-test-utils",
+ "qp-rusty-crystals-threshold",
  "rand",
  "zeroize",
 ]
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ use qp_rusty_crystals_hdwallet::{generate_mnemonic, HDLattice};
 let mut seed = [0u8; 32];
 getrandom::getrandom(&mut seed).expect("Failed to generate seed");
 
-// Generate mnemonic
+// Generate mnemonic (returned as a self-wiping Zeroizing<String>)
 let mnemonic = generate_mnemonic((&mut seed).into())?;
 
 // Create HD wallet
--- a/dilithium/README.md
+++ b/dilithium/README.md
@@ -71,7 +71,7 @@ assert!(is_valid);
 ### Hedged Signing (Deterministic with Entropy)
 
 ```rust
-use qp_rusty_crystals_dilithium::ml_dsa_87;
+use qp_rusty_crystals_dilithium::{ml_dsa_87, SensitiveBytes32};
 
 let mut entropy = [0u8; 32];
 getrandom::getrandom(&mut entropy).expect("Failed to generate entropy");
@@ -81,9 +81,12 @@ let message = b"Message to sign";
 
 let mut hedge_entropy = [0u8; 32];
 getrandom::getrandom(&mut hedge_entropy).expect("Failed to generate hedge entropy");
+// Move the hedge into a self-wiping wrapper (this zeroes `hedge_entropy`);
+// `sign` borrows it, so no unwiped copy of the randomizer is left behind.
+let hedge = SensitiveBytes32::new(&mut hedge_entropy);
 
 let signature = keypair
-	.sign(message, None, Some(hedge_entropy))
+	.sign(message, None, Some(&hedge))
 	.expect("Signing should succeed");
 let is_valid = keypair.verify(message, signature.as_slice(), None);
 assert!(is_valid);
--- a/dilithium/src/acvp.rs
+++ b/dilithium/src/acvp.rs
@@ -122,10 +122,14 @@ macro_rules! acvp_for {
 						let message = hexs(test, "message");
 						let sk = fixed::<SECRETKEYBYTES>(&hexs(test, "sk"), "sk");
 						let expected_sig = hexs(test, "signature");
+						// Move the vector's rnd into a self-wiping wrapper:
+						// `signature_var` borrows the hedge instead of taking
+						// it by value (see `prepare_signing_context`).
 						let hedge = if deterministic {
 							None
 						} else {
-							Some(fixed::<32>(&hexs(test, "rnd"), "rnd"))
+							let mut rnd = fixed::<32>(&hexs(test, "rnd"), "rnd");
+							Some(crate::SensitiveBytes32::new(&mut rnd))
 						};
 
 						let mut sig = [0u8; SIGNBYTES];
@@ -144,7 +148,7 @@ macro_rules! acvp_for {
 							PUBLICKEYBYTES,
 							SECRETKEYBYTES,
 							SIGNBYTES,
-						>(&mut sig, &[], &message, &sk, hedge);
+						>(&mut sig, &[], &message, &sk, hedge.as_ref());
 
 						assert_eq!(
 							sig.as_slice(),
--- a/dilithium/src/frontend.rs
+++ b/dilithium/src/frontend.rs
@@ -272,11 +272,19 @@ macro_rules! define_ml_dsa {
 			///   byte-reproducible signatures are required, especially when key-blob storage
 			///   integrity cannot be guaranteed (see [`SecretKey::from_bytes`] on the unvalidatable
 			///   `K`).
+			///
+			/// The hedge is taken as a borrowed `SensitiveBytes32` rather than bare
+			/// bytes (security review): building the wrapper (via `SensitiveBytes32::new`)
+			/// wipes the caller's raw buffer, the wrapper zeroizes itself on drop, and
+			/// signing only ever *borrows* the bytes — so no unwiped `Copy` of `rnd` is
+			/// smeared through parameter slots on the call chain. A recovered `rnd` plus
+			/// a compromised `K` reconstructs ρ' and enables the known-mask attack
+			/// hedging is meant to prevent.
 			pub fn sign(
 				&self,
 				msg: &[u8],
 				ctx: Option<&[u8]>,
-				hedge: Option<[u8; SEEDBYTES]>,
+				hedge: Option<&$crate::SensitiveBytes32>,
 			) -> Result<Signature, SignatureError> {
 				self.secret.sign(msg, ctx, hedge)
 			}
@@ -363,12 +371,13 @@ macro_rules! define_ml_dsa {
 			/// Compute a signature for a given message.
 			///
 			/// See [`Keypair::sign`] for the argument contract, in particular the
-			/// deterministic-vs-hedged trade-off of `hedge`.
+			/// deterministic-vs-hedged trade-off of `hedge` and why it is passed
+			/// as a borrowed self-wiping wrapper.
 			pub fn sign(
 				&self,
 				msg: &[u8],
 				ctx: Option<&[u8]>,
-				hedge: Option<[u8; SEEDBYTES]>,
+				hedge: Option<&$crate::SensitiveBytes32>,
 			) -> Result<Signature, SignatureError> {
 				if msg.len() > MAX_MESSAGE_SIZE {
 					return Err(SignatureError::MessageTooLong);
@@ -499,8 +508,8 @@ macro_rules! basic_variant_tests {
 			let msg = b"ctx";
 			let h1 = basic_test_entropy();
 			let h2 = basic_test_entropy();
-			let s1 = keys.sign(msg, Some(b"a"), Some(h1.0)).unwrap();
-			let s2 = keys.sign(msg, Some(b"a"), Some(h2.0)).unwrap();
+			let s1 = keys.sign(msg, Some(b"a"), Some(&h1)).unwrap();
+			let s2 = keys.sign(msg, Some(b"a"), Some(&h2)).unwrap();
 			assert_ne!(s1, s2);
 			assert!(keys.verify(msg, &s1, Some(b"a")));
 			assert!(!keys.verify(msg, &s1, Some(b"b")));
--- a/dilithium/src/ml_dsa_87.rs
+++ b/dilithium/src/ml_dsa_87.rs
@@ -32,7 +32,7 @@ mod tests {
 		let mut entropy = get_random_bytes();
 		let keys = Keypair::generate(&mut entropy);
 		let hedge = get_random_bytes();
-		let sig = keys.sign(&msg, None, Some(hedge.0)).unwrap();
+		let sig = keys.sign(&msg, None, Some(&hedge)).unwrap();
 		assert!(keys.verify(&msg, &sig, None));
 	}
 
@@ -42,7 +42,7 @@ mod tests {
 		let mut entropy = get_random_bytes();
 		let keys = Keypair::generate(&mut entropy);
 		let hedge = get_random_bytes();
-		let sig = keys.sign(&msg, None, Some(hedge.0)).unwrap();
+		let sig = keys.sign(&msg, None, Some(&hedge)).unwrap();
 		assert!(keys.verify(&msg, &sig, None));
 	}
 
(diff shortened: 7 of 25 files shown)


n13 (approved):
## Review — verdict: **Approve**

Security hardening across all three crates, with an unusually disciplined methodology: every reproducible finding ships a regression test that is demonstrably red before the fix and green after. I independently re-verified the load-bearing claims rather than taking the PR description at face value.

### Independent verification

- **Red → green confirmed, not just claimed.** I ran the two new heap-scanning tests against `master` (eb71423): `generated_mnemonic_never_survives_in_freed_heap_memory` and `pending_privates_queue_growth_never_frees_unwiped_ks` both **fail** there, and both pass on this branch. The leaks were real, and the fixes eliminate them.
- **Full workspace suite green** on this branch (`cargo test --workspace --all-features`), including doctests — the three new `compile_fail` doctests on `ResharingSignerConfig`, `ResharingRound4Message`, and `NewShareData` behave as intended.
- **Release-mode painted-stack probe** (`sign_stack_zeroization`) passes, covering the new hedged-signing scenario.
- **Mnemonic custody chain holds end to end.** bip39 2.2.2's `Mnemonic` stores only word indices (`[u16; 24]`) — no phrase string, no retained entropy — and derives `ZeroizeOnDrop` under the workspace-enabled `zeroize` feature. `normalize_utf8_cow` keeps the `Cow` borrowed for already-normalized ASCII phrases, so `mnemonic_to_seed`'s parse path allocates nothing hidden. The phrase's only heap allocation is the `join` result the `Zeroizing<String>` owns and wipes, and `mnemonic_to_seed` wraps its incoming `String` in `Zeroizing` on entry. The "single allocation, owned and wiped" claim checks out.
- **Hedge plumbing is sound.** `rnd` is now borrow-only end to end; `&ZERO_RND` in deterministic mode is a promoted const absorbing public zeros, and `derive_mask_seed` already took references — so signatures stay byte-identical (consistent with the unchanged ACVP/KAT results). The `Option<&SensitiveBytes32>` shape matches the `Keypair::generate` convention and removes the API-custody gap regardless of whether the old copies were observable in dead stack memory on a given target.
- **B-tree node-split rationale verified.** `MAX_PARTIES = 6` (error.rs:319), worst case is a party in C(5,3) = 10 subsets → ≤10 entries, inside one 11-entry node. The defense-in-depth boxing with the documented invariant is accurate.
- **Resharing opacity is coherent.** The legitimate wire boundary is serialized `Zeroizing<Vec<u8>>` via `Action::SendPrivate`, so `pub(crate)` fields don't impede downstream use; the `internal-test-helpers` self-dev-dependency correctly keeps forge/tamper access out of production builds (dev-deps don't leak features downstream, path-only dev-deps are stripped on publish). The stated caveat (public Borsh impls still allow determined extraction) is honest and correctly framed as closing the accidental path, not a hard boundary.
- **First-message-wins assessment is correct.** Under the documented sender-authenticated transport, an attacker can only poison its own slot — no stronger than withholding, with attributed abort — while replace-on-duplicate would enable adaptive equivocation. Keeping the policy, documenting the argument, warning on conflicting accept signatures, and pinning it with `duplicate_round4_and_accept_messages_never_displace_the_first` is the right outcome. The `handle_accept_message` change is behavior-preserving plus logging.
- Benches/examples only ever pass `None` hedges, so the `sign` signature change doesn't break them; README and KAT/ACVP call sites were updated consistently.

### Notes (non-blocking)

1. **Versioning — action needed at release time, not in this PR.** All three crates carry versions that are *already tagged with the old APIs* (`dilithium-v4.0.0`, `hdwallet-v4.0.0`, `threshold-v3.0.0`, all tagged 2026-07-31), and this PR breaks all three public APIs (`sign` hedge type, `generate_mnemonic` return type, resharing field visibility). The next release proposals must select **major** bumps (dilithium 5.0.0 / hdwallet 5.0.0 / threshold 4.0.0). The release-proposal workflows take `version_type` as an input, so this is a process reminder for whoever triggers the release.
2. **Cosmetic:** the new comment on `buffer_round1_private` says "A rejected duplicate drops the box, wiping it" — for a duplicate, `or_insert_with` never invokes the closure, so no box is created; the *unboxed* `msg` is dropped and wiped via `ZeroizeOnDrop`. The mechanism is right; the wording describes it slightly wrong.
3. **Pre-existing, not introduced here:** the root README example (`README.md:77-90`) uses `HDLattice::from_mnemonic`, which no longer exists in the crate, and pins `qp-rusty-crystals-hdwallet = "1.0.0"`. This PR touched that code block, so a follow-up fixing the example would be welcome.
4. **Acknowledged residuals are honestly documented** and appropriately scoped as follow-ups: the by-value clone through `poke` → `serialize_round1_private` can leave K_S temporaries in dead stack frames (heap is covered, stack is not), and upstream bip39's `from_entropy_in` holds entropy-derived bits in an unwiped stack array. Neither is made worse by this PR.

### Summary

Real leaks, correctly diagnosed (the `Vec` realloc / B-tree split custody-gap analysis is exactly right), minimal and well-scoped fixes, regression tests that provably bite, and PR description claims that survived independent re-verification. Ship it — with a major version bump at the next release of each crate.

n13 (approved):
## Review

Strong PR. Every fix is narrowly scoped, the custody reasoning in the comments is accurate, and the red-test-first discipline (heap-scanning allocators, painted-stack probes, `compile_fail` pins) gives real evidence rather than assertions. The honest disclosure that the hedged stack probe did *not* fail against the old code is appreciated — it correctly frames that fix as removing a codegen-dependent hazard, not a demonstrated leak. All CI checks are green, including the full workspace suite on two platforms, no_std, and the stack budget.

Things I verified directly:

- **Deterministic signing is byte-identical**: `prepare_signing_context` now absorbs a static `ZERO_RND` where the old code absorbed `unwrap_or([0u8; 32])` — same bytes on the deterministic path, and the hedged path only changes custody, not values. ACVP/KAT staying green confirms it.
- **The `#[should_panic(expected = "invntt_tomont precondition violated")]` test** matches the actual `debug_assert` message in `poly.rs:236`, and the corrected doc bound (`|c| < Q`) matches the per-poly contract.
- **The new DKG realloc test is sound**: `Dkg::message` wraps the incoming `Vec<u8>` in `Zeroizing` immediately (`protocol.rs:700-706`), so the test harness's plain-`Vec` queue handoff is wiped by the callee and can't trip the scanner; the reference-run pattern lives in a never-freed static; one test per file avoids scanner races.
- **The `compile_fail` doctest import paths all resolve** (`resharing::{ResharingSignerConfig, TranscriptSigner, ResharingRound4Message, NewShareData}` are all re-exported), so today they fail on the private-field error, not a broken import. Ran `cargo test -p qp-rusty-crystals-threshold --doc` locally: green.
- **The self-referential dev-dependency** is path-only (no `version`), so it is stripped on publish as the comment claims. The feature leaks into other workspace members' *test* builds via unification, which is fine; production builds never see it.
- `buffer_round1_private`'s `or_insert_with` correctly drops (and thereby wipes) a rejected duplicate, and `pop_pending_private`'s clone-then-drop-box keeps the heap copy wiped in place. The remaining stack-residual is documented as such, consistent with the crate's stated threat model.

### Follow-up: the same B-tree hazard class is live in resharing, at larger scale

Pre-existing code, not touched by this PR, so not a blocker here — but it is the direct sibling of what this PR fixes, and unlike the DKG map it does not sit safely below the node-split threshold:

- `my_subshares: BTreeMap<SubsetPair, NewShareData>` (`resharing/protocol.rs:2012`), the per-recipient `by_recipient` maps, and each `ResharingRound4Message.contributions` all hold `NewShareData` (~16 KB of secret coefficients for ML-DSA-87) **inline** in B-tree nodes.
- If my subset math is right, these maps exceed the 11-entry node capacity at already-supported sizes: for 2-of-5, subsets have size 4 and C(5,4) = 5, so party 0 (lowest-id member, hence designated dealer of every subset containing it) deals for 4 old subsets × 5 new subsets = 20 `my_subshares` entries, and a single recipient's `contributions` map reaches 4 × 4 = 16 entries. Node splits then copy 16 KB secret structs bytewise and strand stale copies exactly as described in the `round1_privates` field doc — where boxing was justified "so the invariant does not silently break if larger committees are ever enabled." In resharing it appears to be broken at sizes supported today.
- `share.clone()` into `by_recipient` (`protocol.rs:2040`) also duplicates the coefficients through a by-value move; probably the documented stack-residual class, but worth folding into the same follow-up.

Suggest a follow-up issue applying the same `Box` treatment (or an explicit size-cap argument if I've miscounted the subset enumeration).

### Before publish (not blocking merge)

All three crates break their public API (`sign` hedge parameter, `generate_mnemonic` return type, resharing field privatization), but the manifests still carry the exact versions already live on crates.io (dilithium 4.0.0, hdwallet 4.0.0, threshold 3.0.0). Majors will be needed at release time — flagging so it's not lost between this merge and the per-crate release PRs. Relatedly, `hdwallet/README.md` still shows `qp-rusty-crystals-hdwallet = "3.1"` in the installation snippet (pre-existing staleness, cheap to fix while the README is being touched).

### Minor

- **hdwallet doesn't re-export `Zeroizing`**, yet `generate_mnemonic` now returns `Zeroizing<String>`. Downstream code that wants to *store* the phrase (struct field, function signature) must add a version-matched `zeroize` dependency just to name the type. The crate already re-exports `SensitiveBytes32`/`SensitiveBytes64`; `pub use zeroize::Zeroizing;` would match that convention.
- The `compile_fail` pins are inherently unpinned to an error *kind* on stable rustdoc: if `resharing::TranscriptSigner` were ever renamed, the doctest would silently pass on the import error instead of the privacy error. Paths are valid today (verified); a one-line comment noting they must stay resolvable would keep the pins honest.

## Verdict

**Approve.** The fixes are correct, behavior-preserving where they must be (ACVP/KAT-pinned), and the regression tests meaningfully raise the bar. The resharing B-tree finding and the semver bumps are follow-ups, not blockers.

n13 (changes requested):
## Review — verdict: **Request changes**

The core API and ownership changes are directionally sound, and all existing checks pass, but two heap-custody gaps remain in the exact secret paths this PR is hardening.

### Blocking findings

1. **[P1] Eliminate the mnemonic word-pointer allocation** (`hdwallet/src/lib.rs:443`)

   `collect::<Vec<&str>>()` allocates a buffer containing the ordered fat pointers for all 24 selected words in the static BIP-39 word list. That pointer sequence is an equivalent encoding of the recovery phrase: a heap reader can map the addresses back to the word list. The temporary Vec is freed immediately after `join` without being wiped, so wrapping only the resulting String does not establish the claimed single secret allocation.

   The added regression scans only for the UTF-8 phrase and cannot see this representation. I reproduced the gap on head with a deallocation-scanning allocator that matches the raw 24-reference sequence; it fails when `generate_mnemonic` frees the temporary Vec. Build the `Zeroizing<String>` directly, using a non-allocating pass to compute exact capacity and a second pass to append words, and extend the regression to cover this representation.

2. **[P1] Keep resharing sub-shares out of inline BTreeMap storage** (`threshold/src/resharing/types.rs:1175`, `threshold/src/resharing/protocol.rs:2062`, `threshold/src/resharing/protocol.rs:2090`)

   `NewShareData` remains stored inline in several BTreeMaps, including `my_subshares` and per-recipient `contributions`. BTreeMap node splits byte-move values and leave stale source slots outside the live length, exactly the mechanism this PR correctly fixes for DKG `round1_privates`. This is reachable at supported sizes: a 3-of-6 to 3-of-6 reshare can give the lowest dealer 150 entries in `my_subshares` and 100 entries in one recipient map, well beyond a single node.

   I reproduced this on head by inserting and dropping 100 distinctive `NewShareData` values under a freed-memory scanner; coefficient bytes remained in a freed node despite `ZeroizeOnDrop` on live values. Store secret-bearing resharing values behind stable allocations, such as `Box<NewShareData>`, across every map that can split, and add a supported-configuration regression test.

### CI coverage

- **[P2] Run the new hedge stack regression in CI.** The debug workspace job compiles `sign_stack_zeroization` to zero tests, while the explicit release job runs only `import_stack_zeroization`. I ran `cargo test --release --locked -p qp-rusty-crystals-dilithium --all-features --test sign_stack_zeroization` manually and all three variants pass, but the new hedged scenario is not currently enforced by CI.

### Validation

- GitHub checks: all green on `1713bcfcc3c635d0a16fd41b9ee4340a4e8a909a`
- `cargo test --workspace --all-features`: pass
- release `sign_stack_zeroization`: 3 passed
- `git diff --check`: pass
- two focused negative heap probes: both reproduce the gaps above

The signing hedge borrowing, DKG boxing, field visibility changes, first-message-wins policy, and inverse-NTT contract update otherwise look coherent.

n13:
## Independent verification of the "Request changes" review

I re-verified all three findings from the latest review against head `1713bcf`, with fresh reproductions written from scratch. **All three are confirmed.** The Request-changes verdict is warranted — the two P1s sit in exactly the custody class this PR is hardening, and both are reachable today.

### 1. Mnemonic word-pointer allocation — **confirmed (P1)**

`generate_mnemonic`'s `mnemonic.words().collect::<Vec<&str>>()` allocates a 384-byte buffer of 24 fat pointers (address + length) into the static BIP-39 word list, freed unwiped immediately after `join`. Since the word-list addresses are fixed statics, the pointer sequence decodes back to the phrase.

I reproduced it independently: a dealloc-scanning allocator that learns the exact fat-pointer byte sequence (by evaluating the same `collect` expression against the same bip39 statics) and scans freed blocks while `generate_mnemonic` runs. **Fails on head** — the freed `Vec` buffer contains the full 24-pointer sequence. The PR's `mnemonic_heap_zeroization` test scans only UTF-8 phrase bytes and cannot see this representation, as the review says.

I also validated the proposed remediation: building the `Zeroizing<String>` directly with an exact-capacity first pass and an append second pass turns both probes green (my fat-pointer probe and the existing phrase probe). One caution for the fix: the capacity computation must be exact (or at least sufficient) — an under-sized `String` reallocates while growing and would strand unwiped partial-phrase prefixes, trading one leak for another.

### 2. `NewShareData` inline in B-tree nodes — **confirmed (P1)**

Reproduced with the crate's real `NewShareData` (via the `internal-test-helpers` constructor): inserting 100 entries with distinctive coefficients into a `BTreeMap<(u16, u16), NewShareData>` and dropping the map trips a freed-memory scanner — stale coefficient bytes survive in freed nodes despite `ZeroizeOnDrop` on every live value. Re-running the identical probe with `Box<NewShareData>` values passes, so the proposed remediation cures it.

The review's reachability numbers also check out from the code: `designated_dealer_for` picks the lowest active member of each subset, so in a 3-of-6 → 3-of-6 reshare party 0 deals for all C(5,3) = 10 old subsets containing it; with C(6,4) = 15 new subsets that is 150 `my_subshares` entries, and one recipient's `contributions` map gets 10 × 10 = 100 — both far past the 11-entry node capacity, so splits are guaranteed at supported sizes. (This is the follow-up my earlier review flagged from the subset math; the reproduction confirms it and I agree with elevating it to blocking — the `round1_privates` boxing rationale in this PR applies verbatim, except here the invariant is already broken today.)

### 3. CI never runs the new hedge stack probe — **confirmed (P2)**

`sign_stack_zeroization.rs` line 36 is `#![cfg(not(debug_assertions))]`, so the debug `cargo test --workspace` job compiles it to zero tests; the only release-mode step in `ci.yml` (lines 100–101) runs `--test import_stack_zeroization` exclusively. No CI job executes the sign/keygen/hedge stack probes. The release step already has the right zero-tests-ran guard — extending it to also run `--test sign_stack_zeroization` (per parameter set, as the job does for features) closes this.

### Verdict

All findings verified as real and correctly prioritized; I concur with **Request changes**. My earlier approval stands superseded on the two P1s — the hedge-borrowing, DKG boxing, visibility, and first-wins work remains sound and none of it needs to be unwound; the fixes here are additive.

n13 (approved):
## Re-review — verdict: **Approve**

Re-reviewed head `bc2bd2d92ab1295516a8f56b1fbc2a9ae407a125`. The three findings from my previous review are resolved:

- Mnemonic generation now wipes the temporary `Vec<&str>` pointer table before deallocation, retains the phrase in `Zeroizing<String>`, and includes a deallocation-scanning regression for the recoverable pointer representation. Both optimized mnemonic probes pass locally.
- Secret-bearing resharing maps now store `NewShareData` behind stable `Box` allocations across the protocol and serialization paths. The supported-size BTreeMap-node regression passes in release mode, and the existing resharing and serialization suites remain green.
- CI now explicitly runs all release-only zeroization probes, including `sign_stack_zeroization`, with a guard against accidentally executing zero tests.

I found no blocking issues in the new commits.

Validation:

- `cargo test --workspace --all-features`: pass
- release mnemonic pointer/stack probes: 3 passed
- release Dilithium signing stack probe: 3 passed
- release resharing BTreeMap-node probe: 1 passed
- `git diff --check`: pass
- GitHub checks: all green on the reviewed head (Ubuntu, macOS, no_std, stack budget, clippy/docs, audit, and CodeQL)

Non-blocking note: the threshold crate no-std configuration already fails on the base branch for unrelated format/math issues. The newly added `Box` references also rely on the std prelude, so `alloc::boxed::Box` should be imported in the changed modules when that existing no-std target is restored; this PR does not regress a previously passing configuration.

PR #115, opened 2026-08-06
Memory hygiene, protocol liveness, and construction-boundary validation

illuzen:
# Security review remediations: memory hygiene, protocol liveness, and construction-boundary validation

This PR addresses twelve security review findings across `dilithium`, `hdwallet`, and `threshold`. Every item followed the same discipline: reproduce the finding with a failing (red) test on the unfixed code, apply the fix, confirm the test turns green, and pin the behavior against regression. Stack-residue findings are verified with release-mode painted-stack probes that scan dead stack memory for secret byte patterns after the operation under test completes.

## Secret zeroization — stack and heap residue

**Wipe DKG subset seeds and stop moving `Round1Private` by value** (`bed97bc`)
Subset seeds derived during DKG contribution computation and commitment verification were left on the stack, and queued `Round1Private` values were moved (i.e. copied) by value through dead stack slots that `ZeroizeOnDrop` can never reach. Seeds are now explicitly wiped after each derivation, the pending private is returned as its original `Box` instead of a clone, and serialization goes through a by-reference `Round1PrivateFrame` wrapper (pinned byte-identical to the owned enum encoding). Targeted unit probes scan the stack immediately after each fixed function.

**Borrow the seed in derivation entrypoints instead of moving it** (`6c04039`)
`derive_key_from_seed` and `generate_wormhole_from_seed` took the 64-byte master seed by value; every call moved an unwipeable copy through the caller's frame, and the mnemonic-based helpers leaked the stretched seed the same way. The entrypoints now borrow (`&SensitiveBytes64`), so the caller's wrapper remains the single owned copy and can be reused across derivations. New probes cover direct-seed and mnemonic-derived paths for both key and wormhole derivation.

**Wipe the DKG master seed once Round 1 derivation consumed it** (`707c06f`)
The DKG kept the master seed in memory for the protocol's entire lifetime although nothing reads it after Round 1 randomness derivation. It is now zeroized immediately after that derivation, shrinking the exposure window from the whole session to a few lines.

**Wipe secrets on rejected `PrivateKeyShare` imports** (`19514ca`)
A malformed or truncated share blob rejected mid-deserialization left the already-parsed `key`, `rho`, `tr`, and share coefficients in temporaries. Deserialization now reads sensitive fields through `Zeroizing` buffers and share entries in place into pre-zeroed map slots with per-coefficient range validation, so rejection at any offset leaves no plaintext copies. A probe exercises both rejection modes (range violation and EOF).

**Keep new-share coefficients off the stack in resharing final assembly** (`355b04c`)
`build_private_key_share` staged `s1`/`s2` coefficient arrays through unwiped stack intermediaries. Coefficients are now copied heap-to-heap into preallocated zeroed map entries, and the stack copy of the assembled party key is wiped after use.

**Wipe the wormhole `first_hash` reveal on pair drop** (`3e51f77`)
`WormholePair` held `first_hash` — the single hash of the salted preimage, a proving reveal, not public data — in a bare `[u8; 32]`, so dropping the pair wiped only the secret. It now uses the same move-only `SensitiveBytes32` wrapper as the secret. Wrapping alone was insufficient: the probe still found one copy, traced to `hash_twice()` recomputing the inner digest inside `qp-poseidon-core`'s frames. The derivation now computes the first hash once, directly into the wrapper, and derives the address from it with `rehash_to_bytes` — byte-identical output (pinned by the golden known-answer test and a new equivalence assertion) with no unwipeable foreign-frame copy.

**Wipe resharing session ephemerals at terminal failure** (`6b2f99b`)
Terminal `Failed` transitions returned without erasing session material, so a failed protocol object retained for diagnostics kept the seed, entropy, session seed, sub-shares, Round 4 buffers, and transcript signer readable until drop. Every failure path now routes through a `fail()` helper that erases the ephemerals at the transition itself. Deliberate deviation from the finding: the old-share working copy is kept on failure — resharing produced no replacement, so the old share is still the party's live key material and must stay recoverable via `abort_and_take_existing_share` (which now wipes the remaining ephemerals right after handing over the share).

## Session binding and protocol liveness

**Bind resharing Round 1 entropy to the session (SSID)** (`02f4ef0`)
Entropy contributions were derived from the constructor seed alone, so a reused seed produced identical, predictable contributions across sessions. The derivation now absorbs the SSID (session nonce, epoch, committees) under a bumped domain tag (`v2`), making contributions session-unique.

**Let callers terminate stalled signing sessions** (`e4eaefc`)
The sans-io, `no_std` signing reactor had no way to end a session stalled by a silent peer — it waited forever. A new `abort_stalled()` transitions any non-terminal state to `Failed` with a `TimedOut` error naming the missing participants, with timeout policy documented as a caller responsibility (consistent with the resharing protocol and the near-mpc integration).

**Make reactor `poke()` errors terminal** (`30cf1b4`)
Errors escaping `poke()` in the signing and resharing reactors could leave non-terminal states, so a supervisor re-polling the instance re-executed the same failing work forever (round inputs are frozen first-write-wins, so the outcome cannot change). Both `poke()` entry points now transition to `Failed` before propagating any error. Attack-surface boundary pinned by tests: `message()` errors stay non-terminal, so a compromised party's junk frame cannot kill a session — missing input is `Action::Wait`, never an error.

## Construction-boundary validation and arithmetic contracts

**Bind the share's rho to the public key at signer construction** (`a0cd5b6`)
`ThresholdSigner::new` compared only `tr`, but response generation expands matrix A from `private_key.rho()` while signature combination reconstructs A from the public key encoding's first 32 bytes. A share with a tampered rho (original tr retained) passed construction and wasted local and peer computation before failing at combination. Construction now rejects the pair unless the share's rho matches the matrix seed embedded in the public key.

**Exclude the endpoint where `montgomery_reduce` returns Q** (`9dbb51f`)
`montgomery_reduce` documented the inclusive domain `[-2^31·Q, 2^31·Q]`, but at exactly `2^31·Q` the signed low word truncates to `i32::MIN` (Q is odd) and the reduction returns exactly `Q` — outside its strict `(-Q, Q)` output contract, violating downstream assumptions (`invntt_tomont`'s `|c| < Q`, canonical packing, norm checks). The domain is now documented and debug-enforced as half-open; the negative endpoint stays inclusive (it reduces to 0). Unreachable by honest protocol data (NTT-domain products stay ~4× under the bound), so the algorithm itself is unchanged.

## Verification

- Release-mode painted-stack probes (new and extended) for every residue finding, each demonstrated red on the unfixed code.
- Golden known-answer tests confirm derivation outputs (wormhole address/first-hash/secret, HD keys) are byte-identical where derivation shapes changed.
- Full workspace suites green: `dilithium` (135 lib + integration), `hdwallet` (51 lib + all probes), `threshold` (288 lib + all integration suites, including the long end-to-end DKG/resharing/signing runs), and the cross-crate integration tests.
- Clippy and nightly rustfmt clean across the touched crates (remaining warnings are pre-existing and unrelated).


n13 (approved):
Fable 5 review 

Reviewed all 13 commits against master, with the claims verified rather than taken from the description. No blocking issues found.

**What was verified**

- **`montgomery_reduce` endpoint math** (`dilithium/src/reduce.rs:11`): confirmed by hand — at `a = 2^31·Q`, `a mod± 2^32 = 2^31` (Q odd) truncates to `i32::MIN`, and since `Q_INV` is odd the low word stays `i32::MIN`, giving `(2^31·Q + 2^31·Q) >> 32 = Q` exactly. The half-open domain, the inclusive negative endpoint, and the tightened `debug_assert` in `pointwise_montgomery` are all correct; release behavior is unchanged and the endpoint is unreachable by honest NTT-domain data.
- **Wormhole derivation-shape change** (`hdwallet/src/wormhole.rs:193`): `rehash_to_bytes(hash_to_bytes(x)) == hash_twice(x)` holds — `hash_to_bytes` emits the canonical digest encoding, `bytes_to_digest` inverts it, so the `.expect()` on canonicality can't fire on library-produced digests. qp-poseidon-core 3.1.0 additionally pins this with its own `test_rehash_equals_hash_twice`, and the golden KAT plus the new equivalence assertion pin it here.
- **`Round1PrivateFrame` wire compatibility** (`threshold/src/keygen/dkg/types.rs:703`): the hardcoded borsh variant index is pinned byte-identical against the owned `DkgMessage::Round1Private` encoding by `round1_private_frame_matches_enum` — reordering the enum fails the test instead of corrupting the wire format. Good defensive design.
- **rho binding** (`threshold/src/signer.rs:252`): ML-DSA public key encoding is `rho ‖ t1`, so comparing `public_key.as_bytes()[..32]` against `private_key.rho()` is the right check, and it closes a real gap (tampered rho + genuine tr passed the old tr-only check).
- **DKG master-seed wipe** (`threshold/src/keygen/dkg/protocol.rs:1065`): grepped all readers of `self.seed` — Round 1 derivation is the only consumer; the later `zeroize()` calls in `Drop`/`abort` become no-ops. The full-DKG test completing with the seed zeroed proves nothing reads it afterward.
- **Terminal-error semantics**: the poke-terminal/message-non-terminal boundary is the right call (a junk frame from a compromised peer must not kill a session; deterministic local failure must not be re-polled forever), and both sides are pinned by tests. `Round4Retry` no longer exists, so the terminal wipes have no retry-path interaction. Keeping the old share on resharing failure (deviation from the finding) is correct — it's still the party's live key material, and `abort_and_take_existing_share` wipes the rest at hand-off.
- **Resharing entropy v2 domain tag** (`threshold/src/resharing/protocol.rs:1213`): SSID binding is local-only input to a commit-reveal, so no wire-format or cross-version interop break; the tag bump is appropriate hygiene.
- **Share-import zeroization** (`threshold/src/keys.rs:221`): the `Zeroizing` header reads and in-place per-coefficient reads are sound; duplicate-mask blobs behave as before (last write wins, canonical-set check still rejects count mismatches).

**Test runs (local, this branch)**

- Release mode, all three crates: **700 passed, 0 failed**, including every new painted-stack probe (`seed_stack_zeroization`, `share_import_stack_zeroization`, the DKG/resharing `stack_probes` modules, `wormhole_secrets_never_survive_pair_drop`) and the new protocol tests (`abort_stalled_*`, `terminal_transitions_wipe_session_secrets`, `entropy_contribution_is_session_bound`, etc.).
- Debug mode: both `should_panic` boundary tests for the montgomery domain pass.

**Non-blocking note**

- **Semver**: `qp-rusty-crystals-hdwallet` 4.0.0 and `qp-rusty-crystals-threshold` 3.0.0 are already published on crates.io, and this PR is breaking for both: `derive_key_from_seed`/`generate_wormhole_from_seed` now take `&SensitiveBytes64` (`hdwallet/src/lib.rs:342`), and `SignProtocolError` gains the `TimedOut` variant on a non-`#[non_exhaustive]` enum (`threshold/src/signing_protocol.rs:284`). Nothing to change in this PR if version bumps happen at release time — just flagging that the next publish of each needs a major bump.

**Verdict: Approve.** Careful, well-tested remediation work — every fix is pinned by a test that was demonstrated red on the unfixed code, the one deliberate deviation from the findings is correctly reasoned, and the equivalence-sensitive changes (wormhole derivation shape, borsh frame) are pinned against drift.

n13 (changes requested):
Request changes: the implementation validates cleanly, but the new release-only security regressions are not wired into CI.

Blocking finding

- **Run every newly added optimized stack probe in CI** (`.github/workflows/ci.yml:114-130`). The normal `cargo test --workspace` job compiles these tests as zero tests because they are guarded by `not(debug_assertions)`, while the explicit release-probe lists omit:
  - `hdwallet/tests/seed_stack_zeroization.rs` (4 tests),
  - `threshold/tests/share_import_stack_zeroization.rs` (1 test), and
  - the `stack_probes` unit modules in `threshold/src/keygen/dkg/protocol.rs` and `threshold/src/resharing/protocol.rs` (4 tests total).

  Consequently, the current green CI run does not exercise the codegen-dependent regressions for borrowed HD seeds, rejected share imports, queued/subset DKG secrets, or resharing final-share assembly. This also conflicts with the workflow's stated safeguard that every optimized probe must run and that zero-test filtering must fail the job. Please add `seed_stack_zeroization` to the HD-wallet loop, `share_import_stack_zeroization` to the threshold integration loop, and a threshold `--lib stack_probes` invocation (or equivalent explicit filters).

Validation performed

- `cargo +nightly fmt --all -- --check` — pass
- `cargo test --workspace --locked` — pass
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` — pass
- `cargo build --target wasm32v1-none -p qp-rusty-crystals-hdwallet --no-default-features --features ml-dsa-44,ml-dsa-65,ml-dsa-87` — pass
- Manually ran all omitted probes above in release mode — 9 passed, 0 failed
- Re-ran the changed `dkg_stack_zeroization` and `wormhole_stack_zeroization` release probes — pass

No other blocking findings were found in the changed implementation.

n13 (approved):
Approve: the follow-up commit fully resolves the previously blocking optimized-probe CI gap.

Reviewed the delta from `b442940` to `e53c110`. `.github/workflows/ci.yml:117-136` now:

- adds `share_import_stack_zeroization` to the threshold release integration probes,
- runs the DKG and resharing release-only unit modules via `--lib stack_probes`, and
- adds `seed_stack_zeroization` to the HD-wallet release probes.

Each invocation still goes through `run_probe`, so the existing nonzero-test guard fails CI if a filter or cfg change silently reduces a probe to zero tests.

Validation:

- Exact local release invocations passed: 1 share-import test, 4 threshold unit stack probes, and 4 HD seed probes.
- `cargo +nightly fmt --all -- --check` passed.
- The updated Ubuntu CI log confirms the same 1 + 4 + 4 tests actually executed in release mode and passed.
- All GitHub checks on head `e53c110` are green, including both workspace matrices, formatting, Clippy/docs, security audit, no-std builds, and stack budget.

No blocking findings remain.

